PR #35, opened 2015-01-28
Fix build on Windows

fanc999:
Hi,

The recent updates to Graphene broke the build on Windows, which can be fixed with some rather simple updates.  I have made the necessary updates to the sources in my fork so that these issues could be remedied.

I will also try to come up with Visual Studio projects/NMake Makefiles to build the sources as soon as possible-sorry I did not have the time to look into Graphene lately due to work on other items.

p.s. Is it correct to assume that there is no intention to support Windows XP for this project, just to double-check (I couldn't check XP compatibility, as I don't normally run it, and I am all for dropping XP support in the GTK+ stack :) ), at least from a thread-safety perspective?

With blessings, thank you!

### DIFF
--- a/src/graphene-private.h
+++ b/src/graphene-private.h
@@ -25,7 +25,6 @@
 #define __GRAPHENE_PRIVATE_H__
 
 #include "config.h"
-#include <stdio.h>
 #include <stdlib.h>
 
 #define GRAPHENE_FLOAT_EPSILON  (1e-15)
@@ -81,6 +80,6 @@
 #define GRAPHENE_DEG_TO_RAD(x)          ((x) * (GRAPHENE_PI / 180.f))
 #define GRAPHENE_RAD_TO_DEG(x)          ((x) * (180.f / GRAPHENE_PI))
 
-#define graphene_lerp(a,b,ff)           (((1.f - (ff)) * (a)) + ((ff) * (b)))
+#define graphene_lerp(a,b,factor)       (((1.f - (factor)) * (a)) + ((factor) * (b)))
 
 #endif /* __GRAPHENE_PRIVATE_H__ */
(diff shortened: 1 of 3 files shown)


ebassi:
thanks for keeping the build up to date!

yes: XP is not going to be supported, and it will be dropped by GTK as well pretty soon.

src/graphene-private.h
```diff
@@ -25,6 +25,7 @@
 #define __GRAPHENE_PRIVATE_H__
 
 #include "config.h"
+#include <stdio.h>
```
ebassi: why do we need stdio.h here?

src/graphene-private.h
```diff
@@ -80,6 +81,6 @@
 #define GRAPHENE_DEG_TO_RAD(x)          ((x) * (GRAPHENE_PI / 180.f))
 #define GRAPHENE_RAD_TO_DEG(x)          ((x) * (180.f / GRAPHENE_PI))
 
-#define graphene_lerp(a,b,f)            (((1.f - (f)) * (a)) + ((f) * (b)))
+#define graphene_lerp(a,b,ff)           (((1.f - (ff)) * (a)) + ((ff) * (b)))
```
ebassi: maybe we can rename the `f` to `factor`.

ebassi:
the fixes look all good; I only have a couple of comments.

src/graphene-private.h
fanc999: Just did that a little bit ago.  Thanks for the notes

src/graphene-private.h
fanc999: Some files are using fprintf() and stderr(), namely graphene-box.c and graphene-vectors.c.  I just moved the stdio.h inclusion into these files directly, if that is better.

ebassi:
looks great! thanks!

PR #40, opened 2015-02-26
Install graphene-config.h in libdir

ndufresne:
This file is architecture dependent. To support multilib we need
to make sure multiple copies can coexist.


ebassi:
CR pass.

PR #47, opened 2016-01-12
Add unit tests for graphene_triangle_get_barycoords()

swilmet:
When reading the API, I thought that u was on the ab axis and v on the ac axis. But it's the reverse. Is that normal?

Also, as the FIXME comment shows in the unit test, getting the barycoords of a 3D point _not_ in the triangle plane returns true (valid coords) nonetheless. Is that normal?

### DIFF
--- a/src/tests/graphene-test-compat.h
+++ b/src/tests/graphene-test-compat.h
@@ -57,6 +57,19 @@
     } \
   } G_STMT_END
 
+#define graphene_assert_fuzzy_vec2_equal(v1,v2,epsilon) \
+  G_STMT_START { \
+    if (graphene_vec2_near (v1, v2, epsilon)) ; \
+    else { \
+      char *s = g_strdup_printf (#v1 " == " #v2 " (+/- " #epsilon "): " \
+                                 "{ x:%.7g, y:%.7g } == { x:%.7g, y:%.7g }", \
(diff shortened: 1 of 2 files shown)


ebassi:
> When reading the API, I thought that u was on the ab axis and v on the ac axis. But it's the reverse. Is that normal?

Unless there's something wrong in the math, which is entirely possible, yes: u is on the AB edge vector (with origin in A), and v is on the AC one, so that if `u + v > 1` it means we crossed the BC edge.

> Also, as the FIXME comment shows in the unit test, getting the barycoords of a 3D point not in the triangle plane returns true (valid coords) nonetheless. Is that normal?

Likely the result of the same bug; it should return false. Though this function is usually employed for hit testing in triangles, which means that the Z coordinate of the point is usually coplanar.

ebassi:
The implementation I followed is [here](http://www.blackpawn.com/texts/pointinpoly/default.html) and it seems to be correct.

ebassi:
Ah, no, my mistake. The implementation on that page puts `v` on the AB vector and `u` on the AC one.

I guess either we swap around the arguments, or we document what `u` and `v` map to.

ebassi:
Let's merge this first, then we can figure out the issue with the non-coplanar point.

PR #51, opened 2016-02-05
Reverse multiplication order for in-place operations

criptych:
This patch reverses the order of multiplication for operations that modify the matrix in place, e.g. rotate, translate, scale.  It _does not_ change the order of the matrix multiply functions, but does add a temp variable and copy (should be negligible, on the order of microseconds or less -- about 15ns on my system).

### DIFF
--- a/src/graphene-simd4x4f.h
+++ b/src/graphene-simd4x4f.h
@@ -431,13 +431,16 @@ graphene_simd4x4f_matrix_mul (const graphene_simd4x4f_t *a,
   const graphene_simd4f_t row3 = a->z;
   const graphene_simd4f_t row4 = a->w;
 
+  graphene_simd4x4f_t r;
+
   /* the order is correct if we want to multiply A with B; remember
    * that matrix multiplication is non-commutative.
    */
-  graphene_simd4x4f_vec4_mul (b, &row1, &res->x);
-  graphene_simd4x4f_vec4_mul (b, &row2, &res->y);
-  graphene_simd4x4f_vec4_mul (b, &row3, &res->z);
-  graphene_simd4x4f_vec4_mul (b, &row4, &res->w);
+  graphene_simd4x4f_vec4_mul (b, &row1, &r.x);
+  graphene_simd4x4f_vec4_mul (b, &row2, &r.y);
+  graphene_simd4x4f_vec4_mul (b, &row3, &r.z);
+  graphene_simd4x4f_vec4_mul (b, &row4, &r.w);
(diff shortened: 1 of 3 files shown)


ebassi:
You really don't explain "why". Saying it brings it closer to OpenGL doesn't mean much, considering that OpenGL does not have any ordering, except in the long since deprecated API.

The order of transformations is completely arbitrary, and I personally prefer the current one. Is there a reason why you need a different order?

src/graphene-simd4x4f.h
```diff
@@ -431,13 +431,16 @@ graphene_simd4x4f_matrix_mul (const graphene_simd4x4f_t *a,
...
    */
-  graphene_simd4x4f_vec4_mul (b, &row1, &res->x);
-  graphene_simd4x4f_vec4_mul (b, &row2, &res->y);
-  graphene_simd4x4f_vec4_mul (b, &row3, &res->z);
-  graphene_simd4x4f_vec4_mul (b, &row4, &res->w);
+  graphene_simd4x4f_vec4_mul (b, &row1, &r.x);
+  graphene_simd4x4f_vec4_mul (b, &row2, &r.y);
+  graphene_simd4x4f_vec4_mul (b, &row3, &r.z);
+  graphene_simd4x4f_vec4_mul (b, &row4, &r.w);
+  res->x = r.x; res->y = r.y; res->z = r.z; res->w = r.w;
```
ebassi: You also conveniently disregard the coding style of Graphene, here.

One statement per line.

criptych:
> Saying it brings it closer to OpenGL doesn't mean much, considering that OpenGL does not have any ordering, except in the long since deprecated API.
> ...
> The usual way to set up drawing in a canvas library is to translate the drawing to a specific position, so that any transformations are relative to that point.

If I understand you correctly, that was the behavior I expected, and the behavior produced by the ordering that OpenGL uses.  When actually using the library I experienced the opposite.  

> Now you require applying all transformations, and then translate, which is backwards. 

I'm talking about arranging it so that *calling* `translate` first actually *applies* it last, "so that any transformations are relative to that point."  This would also simplify using a transform hierarchy, where "local" transforms (translate, rotate, scale) can be applied without regard to what "global" transforms are in effect.

> You also conveniently disregard the coding style of Graphene, here.
> One statement per line.

Sorry, I'll fix that.  

I also just noticed your speed measurements in the comments; looks like the extra 15ns actually brings the speed back down to the "unoptimized" version... :( I'll see if there's a way to remedy that, I suppose I was a bit too quick about submitting this.

ebassi:
This is very unlikely to change, at this point.

PR #52, opened 2016-02-10
tests: add unit test for graphene_triangle_get_area()

swilmet:
The function always returns a positive area. Signed areas can be useful
for some computations in computer graphics, but it is not present in
graphene.


ebassi:
Looks good, thanks!

PR #53, opened 2016-02-10
docs: improve doc of graphene_sphere_translate()

swilmet:
I know what is a translation, but it was not clear to me whether @point
referred to the new center coordinates, or the delta. It turns out that
it's the delta, as expected for a translation, but it's better to
explain it in the docs.


ebassi:
Looks good, thanks!

PR #57, opened 2016-03-30
Fix compile warnings when debug is disabled

cubicool:
Initialize the 'res' pointer to NULL.

Borrow the 'errno' variable rather than using an 'err' variable local
to graphene_aligned_alloc. There is some precedent for this, as the
existing code also set 'errno = 0' before calling the corresponding
alloc routine.


ebassi:
Looks like a worthy simplification as well. Thanks!

PR #60, opened 2016-04-14
Enable SSE codepath on older 32-bit compilers

fanc999:
Hi,

I understand that this may be a rather controversial pull request here. 

For doing graphene_simd4x4f_init() on SSE codepaths, we are passing in 4 __m128's, which is fine on x64 builds as well as newer 32-bit compilers that have special considerations to allow such usage in the ABI, but will fail to build on older 32-bit compilers as a limit of passing 3 __m128's by value is imposed.  This is the case for pre-2012 Visual Studio builds, and possibly older GCC releases, which I unfortunately am not that well acquainted with.

In order to keep ABI disruption to the minimum, we use a check macro to enable such codepaths on 32-bit compilers, and use another macro to get the address of the __m128's if necessary.

I am, by the way, currently only able to deduce the conditions of activating the check macro on Visual Studio builds, not the GCC builds, which I think I might need pointers on, which is not in this PR as of yet.

With blessings, thank you!

### DIFF
--- a/src/graphene-matrix.c
+++ b/src/graphene-matrix.c
@@ -183,10 +183,16 @@ graphene_matrix_init_from_vec4 (graphene_matrix_t     *m,
                                 const graphene_vec4_t *v2,
                                 const graphene_vec4_t *v3)
 {
-  m->value = graphene_simd4x4f_init (v0->value,
-                                     v1->value,
-                                     v2->value,
-                                     v3->value);
+  graphene_simd4f_t p_x, p_y, p_z, p_w;
+
+  p_x = v0->value;
+  p_y = v1->value;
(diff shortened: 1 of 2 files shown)


ebassi:
I'm not really fond of this, especially because I'm not entirely sure I want to support pre-2012 MSVC compilers on 32 bit architectures, if it comes at the expense of maintainability, and at the cost of breaking the API. I guess it's too late, now.

To be fair, I'm perfectly comfortable with the idea of requiring MSVC 2015 or newer on 64 and 32 bit platforms.

fanc999:
Hi Emmanuele,

I see.  Thanks though.

With blessings.

PR #64, opened 2016-06-10
Fixing build errors on ARM NEON.

zdydek:
I was trying to build this for an ARM platform, but ran into some problems.  Not sure if my changes are the best way to resolve them, but the code does compile now.

### DIFF
--- a/src/graphene-matrix.c
+++ b/src/graphene-matrix.c
@@ -746,7 +746,25 @@ graphene_matrix_get_value (const graphene_matrix_t *m,
       return 0.f;
     }
 
-  return graphene_simd4f_get (r, col);
+  switch (col)
+    {
+    case 0:
+      return graphene_simd4f_get (r, 0);
+
+    case 1:
+      return graphene_simd4f_get (r, 1);
+
+    case 2:
+      return graphene_simd4f_get (r, 2);
+
+    case 3:
+      return graphene_simd4f_get (r, 3);
+
+    default:
+      return 0.f;
+    }
+
+  return 0.f;
 }
 
 /**
--- a/src/graphene-simd4f.c
+++ b/src/graphene-simd4f.c
@@ -216,7 +216,19 @@ float
 (graphene_simd4f_get) (const graphene_simd4f_t s,
                        unsigned int            i)
 {
-  return graphene_simd4f_get (s, i);
+  switch (i)
+    {
+    case 0:
+      return graphene_simd4f_get (s, 0);
+    case 1:
+      return graphene_simd4f_get (s, 1);
+    case 2:
+      return graphene_simd4f_get (s, 2);
+    case 3:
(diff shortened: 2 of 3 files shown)


ebassi:
Thanks! I'm setting up an ARM device to test out the changes (and maybe have some sort of automated regression testing), but in principle they look good to me…

src/graphene-matrix.c
```diff
@@ -746,7 +746,25 @@ graphene_matrix_get_value (const graphene_matrix_t *m,
       return 0.f;
     }
 
-  return graphene_simd4f_get (r, col);
+  switch (col)
```
ebassi: Since this has nothing to do with ARM NEON issues, could you split this chunk into its own separate PR?

src/graphene-simd4f.c
```diff
@@ -216,7 +216,19 @@ float
 (graphene_simd4f_get) (const graphene_simd4f_t s,
                        unsigned int            i)
 {
-  return graphene_simd4f_get (s, i);
+  switch (i)
```
ebassi: Same as above, this should go into a separate PR.

src/graphene-matrix.c
zdydek: It does have something to do with ARM NEON, though it is a bit obfuscated.  The underlying ARM fast type that graphene_simd4f_get calls into (I forget which one) requires its inputs to be compile time constants.  Passing the "col" argument in was causing the build to fail.

If you still want me to break out into a separate commit I can do that.

src/graphene-simd4f.c
zdydek: Ditto.

src/graphene-matrix.c
ebassi: Ah, that would have been clearer in a separate commit indeed. ;-)

Yes, if you could split this change in a separate commit and note why it's necessary in the commit log it would be very much welcome — especially because it'll avoid me getting rid of it in the future when I've forgotten about this issue.

james-huang:
hi guys, I've independently some across several of the same issues trying to build arm.  
[commit](https://github.com/james-huang/graphene/commit/11fd3145fcf745109cee514d3d0483be1ce5a36e)

although instead of vreinterpretq_u8_u32 i went with vcgeq_u8.
is vreinterpretq_u8_u32 android specific? 

[ARM Compiler toolchain Compiler Reference] (http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0491f/BABGDDDH.html)

ebassi:
I guess we can close this.

ebassi:
@james-huang I've commented on #65. Looking for `vreinterpretq_u8_u32`, I seem to find it in various toolchains for the NEON set in the arm-eabi, so I'm not sure it's Android specific.

PR #69, opened 2016-09-01
simd4f: Fix a compilation error

ford-prefect:
Not sure why the missing semicolon didn't get caught before.


ebassi:
Thanks for the patch! It looks good, but would it be possible to update the commit message to list:

 * the compiler and version that generated the warning
 * the actual warning

?

I haven't had any warning on my GCC 6.1.1 and Travis didn't find any issue with GCC 4.8 and Clang 3.4, but I don't doubt that there may be combinations that trip the issue. It would be nice to document those.

ford-prefect:
Done now. This was an ARM64 build for Android.

ebassi:
Awesome, thanks!

PR #70, opened 2016-09-20
Fix building graphene as a meson submodule

thiblahute:
Fixes building graphene as a meson submodule

Proposed changes:

   - meson: Fix library version and soversion
   - meson: Do not use add_global_arguments but instead pass the args ourself


ebassi:
Looks good.

PR #73, opened 2016-10-20
build: Allow proper build with Visual Studio out-of-the-box

fanc999:
Hi,

This attempts to add proper support out-of-the box for building Graphene using Visual Studio, so that we can make use of it with Visual Studio builds of the upcoming GTK+-4.x (for GSK), as most of the code is ready for building and usage under Visual Studio 2008+ (!), but a build system that would support the MSVC toolchain is not yet in place.

As a result, updates are needed for the meson build files to support the toolchain properly, and a few minor changes are needed for the code so that things will work in the Visual Studio versions that we support for GTK+ (the code changes I will post later).  Note that on 32-bit builds, Visual Studio 2013 or later is required to enable SSE2 support due to x86/32-bit ABI limitations.

With blessings, thank you!

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -58,35 +58,63 @@ conf.set('GRAPHENE_MICRO_VERSION', graphene_micro_version)
 conf.set('GRAPHENE_VERSION', graphene_version)
 
 # Compiler flags
-test_cflags = [
-  '-ffast-math',
-  '-fstrict-aliasing',
-  '-Wall',
-  '-Wpointer-arith',
-  '-Wmissing-declarations',
-  '-Wformat=2',
-  '-Wstrict-prototypes',
-  '-Wmissing-prototypes',
-  '-Wnested-externs',
-  '-Wold-style-definition',
-  '-Wdeclaration-after-statement',
-  '-Wunused',
-  '-Wuninitialized',
-  '-Wshadow',
-  '-Wmissing-noreturn',
-  '-Wmissing-format-attribute',
-  '-Wredundant-decls',
-  '-Wlogical-op',
-  '-Wcast-align',
-  '-Wno-unused-local-typedefs',
-  '-Werror=implicit',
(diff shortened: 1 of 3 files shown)


src/meson.build
```diff
@@ -89,6 +89,15 @@ platform_deps = []
 if build_gobject
   pkgconfig_files += [ 'graphene-gobject-@0@'.format(graphene_api_version) ]
   platform_deps += [ gobject ]
+  if cc.get_id() == 'msvc'
+    platform_deps += [ glib ]
```
ebassi: Shouldn't gobject bring glib along?

fanc999: Hi @ebassi, 

Unfortunately no, only using pkg-config does that.

meson.build
```diff
@@ -162,8 +190,25 @@ threadlib = dependency('threads', required: true)
 # Optional dependency on GObject
 build_gobject = false
 if get_option('enable-gobject-types')
-  gobject = dependency('gobject-2.0', version: '>= 2.30.0', required: false)
-  build_gobject = gobject.found()
+  # MSVC builds typically do not use pkg-config, so find GObject/GLib manually
```
fanc999: Hmm, I see...

This follows along the Visual Studio build files we currently have in the GTK+ stack, as no pkg-config is used in those, except when doing introspection builds, which this patch does not cover (yet), as the introspection scanner will make use of pkg-config.

TingPing: In the long term I think depending on pkg-config on Windows is only a good thing. It isn't a massive dependency, is required by other layers of the stack (gobject-introspection), and generally makes building easier with build systems such as Meson.

fanc999:
Hi,

Updated the files to go back to using pkg-config for all builds, plus some small changes for pre-2015 Visual Studio.

With blessings, thank you!

PR #78, opened 2016-11-22
build: install tests into libexecdir

ignatenkobrain:
Tests are ELF files, so in case of multilib they will conflict. Using libexecdir will tell that this package is not multilib.

@ebassi though I'm not sure if we should use $libdir instead of $libexecdir.


ignatenkobrain:
eog, glib2, gjs are installing elfs into libexecdir

ebassi:
I've merged manually, and fixed @TingPing's comment as well.

PR #79, opened 2016-11-22
Fix build of 1.5.1 on Visual Studio

fanc999:
Hi,

We have started to use C99-style struct initializers, which is support by Visual Studio 2013/2015, with one exception, that is, using them in initializing static variables and global constants, which will cause C2099 to be raised (initializer not a constant).  This might also affect other compilers as well.

Also, we need to update config.h.meson to define inline as __inline for pre-2013 Visual Studio as well, in addition to graphene-config.h.meson.  I am not sure whether meson has a mechanism for detecting the 'inline' definition for the compiler, so this is what I have for now.

With bessings, thank you!

### DIFF
--- a/src/graphene-point.c
+++ b/src/graphene-point.c
@@ -295,7 +295,11 @@ graphene_point_to_vec2 (const graphene_point_t *p,
   v->value = graphene_simd4f_init (p->x, p->y, 0.f, 0.f);
 }
 
+#ifdef _MSC_VER
+static const graphene_point_t _graphene_point_zero = {0.f, 0.f};
+#else
 static const graphene_point_t _graphene_point_zero = GRAPHENE_POINT_INIT_ZERO;
+#endif
 
 /**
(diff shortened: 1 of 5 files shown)


src/graphene-point.c
```diff
@@ -295,7 +295,11 @@ graphene_point_to_vec2 (const graphene_point_t *p,
   v->value = graphene_simd4f_init (p->x, p->y, 0.f, 0.f);
 }
 
+#ifdef _MSC_VER
+static const graphene_point_t _graphene_point_zero = {0.f, 0.f};
+#else
 static const graphene_point_t _graphene_point_zero = GRAPHENE_POINT_INIT_ZERO;
```
ebassi: This is kind of weird, considering that the macro expands to the same thing.

I assume it's the `(type)` cast annotation that breaks it and makes the compiler think this is not a constant assignment.

PR #84, opened 2016-12-02
meson: Detect and use _aligned_malloc on MinGW

### LINKED ISSUE #83: No aligned malloc implementation on MinGW
When building on Windows with MinGW, there's no available aligned malloc implementation:

configure:
```
checking for aligned_alloc... no
checking for posix_memalign... no
checking for memalign... no
```

meson:
```
Checking for function "aligned_alloc": NO
Checking for function "memalign": NO
Checking for function "posix_memalign": NO
```

This means that `alloc_align()` in `src/bench/matrix.c` ends up returning an uninitialized pointer. The implementation needs to be like this to actually show the bug at compile-time:

```
diff --git a/src/bench/matrix.c b/src/bench/matrix.c
index 2621243..d2d6a0a 100644
--- a/src/bench/matrix.c
+++ b/src/bench/matrix.c
@@ -58,6 +58,8 @@ alloc_align (gsize n,
   res = aligned_alloc (alignment, real_size);
 #elif defined(HAVE_MEMALIGN)
   res = memalign (alignment, real_size);
+#else
+#error "Need some type of aligned allocation function"
 #endif

   g_assert (res != NULL);
```

I noticed this because of this warning:

```
[40/61] Compiling c object src/bench/matrix-scalar@exe/matrix.c.obj
../src/bench/matrix.c: In function 'alloc_align':
../src/bench/matrix.c:50:9: warning: unused variable 'real_size' [-Wunused-variable]
../src/bench/matrix.c:63:3: warning: 'res' is used uninitialized in this function [-Wuninitialized]
```

Maybe graphene should be built with `werror=true` in `meson.build:project()` so that `-Werror` is enabled, or perhaps that option can be enabled with a version check inside `meson.build`.

nirbheek:
Fixes #83 

MinGW also uses `_aligned_malloc()`, not just MSVC. Without this, we were just returning uninitialized memory since no other aligned allocator was available.

Also, use `G_GINT64_FORMAT` for printing `gint64` otherwise we get a segfault while printing.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -130,6 +130,7 @@ conf.set('HAVE_MEMORY_H', cc.has_header('memory.h'))
 
 # Functions
 conf.set('HAVE_ALIGNED_ALLOC', cc.has_function('aligned_alloc', prefix: '#include <stdlib.h>'))
+conf.set('HAVE__ALIGNED_MALLOC', cc.has_function('_aligned_malloc', prefix: '#include <stdlib.h>'))
 conf.set('HAVE_MEMALIGN', cc.has_function('memalign', prefix: '#include <stdlib.h>\n#include <malloc.h>'))
 conf.set('HAVE_POSIX_MEMALIGN', cc.has_function('posix_memalign', prefix: '#include <stdlib.h>'))
 conf.set('HAVE_SINCOSF', cc.has_function('sincosf', prefix: '#define _GNU_SOURCE\n#include <math.h>'))
--- a/src/bench/graphene-bench-utils.c
+++ b/src/bench/graphene-bench-utils.c
@@ -277,7 +277,7 @@ graphene_bench_print_results (const char *impl,
   switch (bench_output)
     {
     case BENCH_FORMAT_NONE:
-      g_print ("### '%s' (%ld iterations - using %s implementation) ###\n"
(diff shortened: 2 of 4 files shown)


ebassi:
Looks good, thanks.

meson.build
```diff
@@ -130,6 +130,7 @@ conf.set('HAVE_MEMORY_H', cc.has_header('memory.h'))
 
 # Functions
 conf.set('HAVE_ALIGNED_ALLOC', cc.has_function('aligned_alloc', prefix: '#include <stdlib.h>'))
+conf.set('HAVE__ALIGNED_MALLOC', cc.has_function('_aligned_malloc', prefix: '#include <stdlib.h>'))
```
achadwick: This needs to be <malloc.h> according to https://msdn.microsoft.com/en-us/library/8z34s9c6.aspx

nirbheek: I thought on MinGW it was only defined in `stdlib.h` so I went with that. It turns out that it is defined in both `malloc.h` and `stdlib.h` for MinGW. Not sure about MSVC, but the safe bet is that it's in `malloc.h`. I guess we should change this. I'll submit a new PR since there's one more change that's needed with this.

achadwick: The argument order is funny too. Putting together an "it compiles" PR now just to be sure it flies on MSYS2.

PR #86, opened 2016-12-05
Use malloc.h for _aligned_malloc instead of stdlib.h

nirbheek:
Fixes https://github.com/ebassi/graphene/pull/84#discussion_r90875361

Both `stdlib.h` and `malloc.h` define it on MinGW but it's only defined by `malloc.h` on MSVC, so just use `malloc.h`.

There are more issues needed for MSVC support, but this should help.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -130,7 +130,7 @@ conf.set('HAVE_MEMORY_H', cc.has_header('memory.h'))
 
 # Functions
 conf.set('HAVE_ALIGNED_ALLOC', cc.has_function('aligned_alloc', prefix: '#include <stdlib.h>'))
-conf.set('HAVE__ALIGNED_MALLOC', cc.has_function('_aligned_malloc', prefix: '#include <stdlib.h>'))
+conf.set('HAVE__ALIGNED_MALLOC', cc.has_function('_aligned_malloc', prefix: '#include <malloc.h>'))
 conf.set('HAVE_MEMALIGN', cc.has_function('memalign', prefix: '#include <stdlib.h>\n#include <malloc.h>'))
 conf.set('HAVE_POSIX_MEMALIGN', cc.has_function('posix_memalign', prefix: '#include <stdlib.h>'))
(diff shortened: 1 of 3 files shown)


achadwick:
Conflicts with #87, which also makes graphene use _aligned_malloc. You cannot assume aligned_alloc() exists if _aligned_malloc is defined.

nirbheek:
> Conflicts with #87, which also makes graphene use _aligned_malloc. You cannot assume aligned_alloc() exists if _aligned_malloc is defined.

I'm confused, did you miss this piece of code in bench/matrix.c?

```
#ifdef HAVE__ALIGNED_MALLOC
/* On Window, we have to use _aligned_malloc() which takes in parameters in
 * inverted order from aligned_alloc(), but things are more or less the same
 * there otherwise
 */
#define aligned_alloc(alignment,size) _aligned_malloc (size, alignment)

/* if we _aligned_malloc()'ed, then we must do _align_free() on MSVC */
#define free_aligned(x) _aligned_free (x)
#else
#define free_aligned(x) free (x)
#endif
```

Graphene defines `aligned_alloc()` as a macro that switches the arguments and calls `_aligned_malloc()` if the latter is being used.

achadwick:
Yes I did, it wasn't there on master. The test in graphene-alloc.c will need to be changed too, 

It looks like your PR is more complete, but the order of the #ifdefs will still need to be changed because posix_memalign is still incorrectly detected.

nirbheek:
I don't think the order of the ifdefs is what the issue is. I think the configure/meson checks should only select and define one aligned memory allocator. Right now it tests for all of them and defines everything it finds.

At the same time, the `posix_memalign` test should just be disabled on Windows since nothing provides it there yet. It looks like MinGW wants to (going by your bug report), but it doesn't yet and it's shipping with a half-finished implementation (fwict) which is why Meson detects it even though it's not fully implemented.

ebassi:
Let's go with this, then.

PR #87, opened 2016-12-05
Use _aligned_malloc

achadwick:
This appears to be needed to get the build a little further on in MSYS2. It is entirely untested.

Proposed changes:

 - Look for _aligned_malloc in the correct place for MSVC and MSYS2 (see ref)
- Actually use _aligned_malloc().
  - used in preference to posix_memalign() because posix_memalign is still incorrectly detected on this system possibly as the result of an incorrectly exposed __builtin_posix_memalign.

Benchmark results:

- Cannot build yet, sorry. Next hurdle for me is the `g-ir-scanner` line, which may be too long for meson to handle.

Test suite changes:

 - Amended src/bench/matrix.c


achadwick:
Withdrawing this one in favour of #86.

PR #93, opened 2017-01-03
Update build docs for MSYS2

achadwick:
> This commit updates README.py with better tested documentation for the
> MSYS2 platform on Windows. Previously the instructions would fail because
> "automake" is not a package.
> 
> This commit also gets the user to select packages based on their intended
> target platform, a concept that matters on MSYS2 because each target
> has its own $PREFIX.
> 
> The docs were also out of date in that they did not cover meson.

--------------------------

Proposed changes:

 - Update README.md as described in the commit message, quoted above.

Benchmark results:

 - I have not been able to get it to build yet, so I have not investigated this.
 - No change is expected because this is a docs-only commit.

Test suite changes:

 - None proposed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ First of all, clone the Git repository:
 Then run:
 
     $ mkdir _build && cd _build
-    $ meson ..
+    $ meson ..    # on Windows, it's "meson.py"
     $ ninja test
     # ninja install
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,7 +63,7 @@ First of all, clone the Git repository:
 Then run:
 
     $ mkdir _build && cd _build
-    $ meson ..
+    $ meson ..    # sometimes it's "meson.py"
```
TingPing: It is only `meson.py` on Windows.

achadwick: I'm not sure I understand. MSYS2 actively eschews any official Windows python on your system: it sets its own PATH. Any meson outside its prefix should be ignored too.

"Sometimes it's "meson.py"" is still accurate :neutral_face:

TingPing: Meson only names the file `meson.py` on Windows, so it isn't a vague "sometimes" it is "on Windows".

And what I meant was Meson probably should treat msys2 special and *not* have the `.py` suffix for it. This is probably something that should be discussed upstream (or msys could even patch that).

achadwick: An MSYS2 runtime cannot be detected with `os.name, sys.platform` for the native "/mingw{32,64}" prefixes of MSYS2 and their Pythons. I think that's by design. Native stuff is supposed to be theoretically able to run without the stuff in the Cygwin "/" prefix, so Python there looks like Windows Python.

So MSYS2 ought to patch away the superfluous .py, or write a compatibility script. It's not really a problem that meson should fix, IMO.

I'll make it "on Windows".

TingPing: > So MSYS2 ought to patch away the superfluous .py, or write a compatibility script. It's not really a problem that meson should fix, IMO.

Would be nice if you could report that to them =)

PR #102, opened 2017-04-07
Updated meson build files to ease API version change.

### LINKED ISSUE #101: Hard coded strings on meson build files
### Experienced behavior

If the package version changes, the hard coded strings have to be changed manually.

### Expected behavior

Proper versioning of files and directories generated when changing the package version number.

### Steps to reproduce

Update project version number on meson build file.

inigomartinez:
meson build files have been changed to include a variable with graphene string to be used instead of hard coded strings.

Fixes #101

Proposed changes:

 - Changed the version number handling for generated files and directories in the meson build files.

Benchmark results:

 - Before: ...
 - After: ...

Test suite changes:

 - ...

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -16,6 +16,8 @@ graphene_micro_version = version_array[2].to_int()
 
 graphene_api_version = '@0@.0'.format(graphene_major_version)
 
+graphene_string = '@0@-@1@'.format(meson.project_name(), graphene_api_version)
+
 # The interface age is reset every time we add new API; this
 # should only happen during development cycles, otherwise the
 # interface age is the same as the micro version
@@ -181,6 +183,7 @@ endif
 # Optional dependency on GObject
 build_gobject = false
 if get_option('enable-gobject-types')
+  graphene_gobject_string = '@0@-gobject-@1@'.format(meson.project_name(), graphene_api_version)
   gobject = dependency('gobject-2.0', version: '>= 2.30.0', required: false)
   build_gobject = gobject.found()
   if build_gobject
--- a/meson.build
+++ b/meson.build
@@ -16,7 +16,7 @@ graphene_micro_version = version_array[2].to_int()
 
 graphene_api_version = '@0@.0'.format(graphene_major_version)
 
-graphene_string = '@0@-@1@'.format(meson.project_name(), graphene_api_version)
+graphene_api_path = '@0@-@1@'.format(meson.project_name(), graphene_api_version)
 
 # The interface age is reset every time we add new API; this
 # should only happen during development cycles, otherwise the
@@ -183,7 +183,7 @@ endif
 # Optional dependency on GObject
 build_gobject = false
 if get_option('enable-gobject-types')
-  graphene_gobject_string = '@0@-gobject-@1@'.format(meson.project_name(), graphene_api_version)
+  graphene_gobject_api_path = '@0@-gobject-@1@'.format(meson.project_name(), graphene_api_version)
   gobject = dependency('gobject-2.0', version: '>= 2.30.0', required: false)
   build_gobject = gobject.found()
   if build_gobject
--- a/src/meson.build
+++ b/src/meson.build
@@ -83,11 +83,11 @@ configure_file(input: 'graphene-version.h.in',
 
 install_headers(headers + simd_headers + [ 'graphene.h' ], subdir: graphene_api_path)
 
-pkgconfig_files = [ graphene_string ]
+pkgconfig_files = [ graphene_api_path ]
 platform_deps = []
 
 if build_gobject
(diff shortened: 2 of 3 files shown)


ebassi:
Thanks for the patch, it looks good and simplifies a bunch of redundant bits.

meson.build
```diff
@@ -16,6 +16,8 @@ graphene_micro_version = version_array[2].to_int()
 
 graphene_api_version = '@0@.0'.format(graphene_major_version)
 
+graphene_string = '@0@-@1@'.format(meson.project_name(), graphene_api_version)
```
ebassi: Maybe this could have a better name; `graphene_string` is a bit opaque.

What about `graphene_api_string`, instead?

meson.build
```diff
@@ -181,6 +183,7 @@ endif
 # Optional dependency on GObject
 build_gobject = false
 if get_option('enable-gobject-types')
+  graphene_gobject_string = '@0@-gobject-@1@'.format(meson.project_name(), graphene_api_version)
```
ebassi: This should be changed to match the variable name above.

meson.build
ebassi: Or `graphene_path`, since it's mostly used for paths and file names.

inigomartinez: One last option, `graphene_api_path`, as it's true that it's mostly used for paths and file names, but it's also related to the api version. What's your opinion?

ebassi: I like `graphene_api_path`, indeed. :+1:

meson.build
inigomartinez: Are you referring to just gobject? Is something like `gobject_api_path` fine?

ebassi: If we use `graphene_api_path` above, then this should be changed into `graphene_gobject_api_path`.

src/meson.build
```diff
@@ -72,22 +72,22 @@ configure_file(input: 'graphene-config.h.meson',
...
                configuration: conf,
                install: true,
-               install_dir: join_paths(graphene_includedir, 'graphene-1.0'))
+               install_dir: join_paths(graphene_includedir, graphene_api_path))
 
-install_headers(headers + simd_headers + [ 'graphene.h' ], subdir: 'graphene-1.0')
+install_headers(headers + simd_headers + [ 'graphene.h' ], subdir: graphene_api_path)
 
-pkgconfig_files = [ 'graphene-@0@'.format(graphene_api_version) ]
+pkgconfig_files = [ graphene_string ]
```
ebassi: Needs to be renamed to `graphene_api_path`.

src/meson.build
```diff
@@ -72,22 +72,22 @@ configure_file(input: 'graphene-config.h.meson',
...
 platform_deps = []
 
 if build_gobject
-  pkgconfig_files += [ 'graphene-gobject-@0@'.format(graphene_api_version) ]
+  pkgconfig_files += [ graphene_gobject_string ]
```
ebassi: Needs to be renamed to `graphene_gobject_api_path`.

src/meson.build
```diff
@@ -100,7 +100,7 @@ endif
 libtype = get_option('default_library')
 
 if libtype != 'static'
-  graphene_shared = shared_library('graphene-@0@'.format(graphene_api_version),
+  graphene_shared = shared_library(graphene_string,
```
ebassi: Needs to be renamed to `graphene_api_path`.

src/tests/meson.build
```diff
@@ -22,8 +22,8 @@ unit_tests = [
 python = python3.find_python()
 gen_installed_test = join_paths(meson.current_source_dir(), 'gen-installed-test.py')
 
-installed_test_datadir = join_paths(get_option('prefix'), get_option('datadir'), 'installed-tests', 'graphene-1.0')
-installed_test_bindir = join_paths(get_option('prefix'), get_option('libexecdir'), 'installed-tests', 'graphene-1.0')
+installed_test_datadir = join_paths(get_option('prefix'), get_option('datadir'), 'installed-tests', graphene_string)
+installed_test_bindir = join_paths(get_option('prefix'), get_option('libexecdir'), 'installed-tests', graphene_string)
```
ebassi: Both need to be renamed to `graphene_api_path`.

inigomartinez:
I'm very sorry about the update regarding your comments. I tried to update them using the edit option, but I haven't found any way to edit multiple files at the same time, so I had to edit them one by one. Also, the web editor search option doesn't work very well here, and I missed some of the variables when updating one of the files :(.

I expect the files to be properly updated now.

ebassi:
@inigomartinez No worries; I've squashed the commits on merge, it's perfectly fine.

Thanks again for your contribution!

Issue #104, opened 2017-05-05
Multiple places for option handling in configure


inigomartinez:
### Experienced behavior

If any option is added to meson, there are multiple places in the configure scripts to be modified to add proper support..

### Expected behavior

Easier way to add an option to meson by adding only the option name and its default value.

### Steps to reproduce

Add a new option to mesa.

ebassi:
I'm sorry, but I don't understand this issue.

What is it that you're trying to achieve?

inigomartinez:
Oh, I was opening the pull request by following the "Contributing" notes. This patch tries to simplify the configure script.

ebassi:
We dropped the Autotools build, and we don't use a build-api wrapper any more.

PR #105, opened 2017-05-05
build: Simplified option handling in configure script (#104)

inigomartinez:
Option handling has been simplified by using associative arrays, where all the options are defined along their default options. Then passed options are analyzed using the "action-option" pattern and added accordingly to the meson command.

Fixes #...

Option handling in configure file.

Proposed changes:

Use of associative arrays and analyze options with the "action-option" pattern.

 - ...

Benchmark results:

 - Before: ...
 - After: ...

Test suite changes:

 - ...

### DIFF
--- a/configure
+++ b/configure
@@ -51,13 +51,33 @@ sanitycheck() {
     exit 1
 }
 
+checkoption() {
+    # $1 = arg
+    option="${1#*--}"
+    action="${option%%-*}"
+    name="${option#*-}"
+    if [ ${default_options[$name]+_} ]; then
+        case "$action" in
+            enable) meson_options[$name]=true;;
+            disable) meson_options[$name]=false;;
+            *) echo -e "\e[1;33mINFO\e[0m: Ignoring unknown action '$action'";;
+        esac
+    else
+        echo -e "\e[1;33mINFO\e[0m: Ignoring unknown option '$option'"
+    fi
+}
+
 sanitycheck MESON 'meson'
 sanitycheck MESONTEST 'mesontest'
 sanitycheck NINJA 'ninja' 'ninja-build'
 
-enable_docs='-Denable-gtk-doc=false'
-enable_man='-Denable-man=false'
-enable_introspection=''
+declare -A default_options=(
+    ['gtk-doc']=false
+    ['man']=false
+    ['introspection']=true
+)
+
+declare -A meson_options
 
 while (($# > 0)); do
     case "${1%%=*}" in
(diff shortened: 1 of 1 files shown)


TingPing:
Might be worth noting that Meson *might* at some point error on invalid options (mesonbuild/meson#932) and that other projects copy paste this `configure` file so automatic mapping could end poorly.

inigomartinez:
While this patch does not point error on invalid options, it shows a warning. It should be rather easy to change the warning for an error and stop the script.

Regarding the use of the configure file in other projects, I've used this script in another project[0], which I've slightly modified with the changes in this pull request. While those changes are related only to the build system and not really significant, I though about them as a minor improvement.

[0] https://bugzilla.gnome.org/show_bug.cgi?id=781908#c10

TingPing:
> While this patch does not point error on invalid options, it shows a warning. It should be rather easy to change the warning for an error and stop the script.

Sorry I was referring to it mapping to Meson options that may not exist and Meson erroring could be a problem.

inigomartinez:
Oh, it's fine, don't worry. There is an associative array which holds the meson options with their default values, and every option used with configure are checked against those values. If the option does exist, it's added as an argument to the meson's command line, but if it doesn't exist it shows a warning (the one I pointed before that could be easily replaced by an error and a stop).

configure
```diff
@@ -51,13 +51,33 @@ sanitycheck() {
...
+}
+
 sanitycheck MESON 'meson'
 sanitycheck MESONTEST 'mesontest'
 sanitycheck NINJA 'ninja' 'ninja-build'
 
-enable_docs='-Denable-gtk-doc=false'
-enable_man='-Denable-man=false'
-enable_introspection=''
+declare -A default_options=(
```
ebassi: I like this idea, but not all options take a boolean value, so maybe `checkoption()` should take the argument as well, like:

    *) opt=$1; read_arg arg "$@" || shift; checkoption $opt $arg ;;

inigomartinez:
You comment has made me rethink the option handling. The last commit presents, more or less, these new ideas.

The option hadling presented in the PR only worked with boolean options by using the _{enable|disable}-{option}_ pattern. It wasn't thought to work with other than boolean options.

I have extended that work to also support other kind of options, as you asked for. That work has led me think about dir options, that could be handled in a similar, but not exactly the same, way.

For example, their default values. I haven't be able to declare their default value in the associative array, because of the relationship between them. For example I can define bindir as _"${dir_options['prefix']}/bin"_, which will be resolved when the array is defined, but will not be updated if the value at the 'prefix' key is updated.

Another issue was related to the array where they are defined. Meson handles in a different way the directory options, so they must be stored in a different array. That is the reason behin _set_option_ and _set_bool_option_. _set_option_ alone does handle both cases at the expense of an extra if statement.

Finally, I tried to use an unified regex to handle all the options, but I have only been able to do so by using two separate regex expression.

ebassi:
The `configure` script is now gone.

PR #108, opened 2017-07-19
Fix Xcode-9 compile error

jralls:
Apple's clang-9 (no idea what that maps to in clang's normal version) doesn't like an '=' in the version flags.

Fixes #...

Proposed changes:

 - ...

Benchmark results:

 - Before: ...
 - After: ...

Test suite changes:

 - ...


jralls:
The error it emits is: ```ld: malformed 32-bit x.y.z version number: =1```

PR #113, opened 2018-03-06
Fix header reference in GIR

ricotz:


ebassi:
Thanks! That'll teach me about doing releases while travelling.

PR #127, opened 2018-10-16
meson: fix HAVE_INIT_ONCE declaration

ystreet:
meson will set the HAVE_INIT_ONCE value to 0 or 1 when using
conf.set10() so an #if defined() will always succeed.

../src/graphene-box.c:681:1: error: unknown type name 'INIT_ONCE'
../src/graphene-box.c:681:36: error: 'INIT_ONCE_STATIC_INIT' undeclared here (not in a function)
../src/graphene-box.c: In function 'init_static_box':


ebassi:
Thanks for your contribution!

To be quite honest, I don't really like this style. Either the `meson.build` should not use `set10()` and instead rely on the definedness of the symbol; or the preprocessor conditional should not be checking for the symbol being defined (as it it will always be) and turned into:

```cpp
#if HAVE_INIT_ONCE
...
```

and:

```cpp
...
#elif HAVE_INIT_ONCE
...
```

PR #128, opened 2018-10-16
Fix implicit conversion error when compiling with MSVC

nirbheek:
error C4244: 'function': conversion from 'int' to 'float', possible loss of data

`i` can't be float because it is also used as an array subscript

This is the only changed needed to have graphene build with msvc.


nirbheek:
I suspect this only shows up with MSVC 19.14.26433, which would explain why the CI didn't trigger it.

PR #130, opened 2018-12-14
bench: Fix MSVC build error on 32-bit Windows

nirbheek:
`size_t` is `unsigned long int` which is a 32-bit integer on 32-bit Windows. The multiplication in `malloc()` below then evaluates to a 32-bit integer which the compiler complains about.

This is normally a warning, but since we use `-we4244` it's an error.

src/bench/graphene-bench-utils.c(175): error C4244: 'function': conversion from 'gint64' to 'size_t', possible loss of data`

The number of rounds will never exceed INT_MAX, so we can just use an int here.


ebassi:
It seems this is failing to build.

PR #131, opened 2018-12-21
build: don't hide the stdbool include from g-ir-scanner

lazka:
g-i recently lost the ability to infer the bool type without a stdbool include.
While this is about to be fixed in g-i (https://gitlab.gnome.org/GNOME/gobject-introspection/merge_requests/116)
I don't see a reason why we should hide this in the first place, so remove the ifdef guard.

### DIFF
--- a/src/graphene-macros.h
+++ b/src/graphene-macros.h
@@ -71,14 +71,12 @@
 # define GRAPHENE_END_DECLS
 #endif
 
-#ifndef __GI_SCANNER__
 #if defined(_MSC_VER) && !defined(__bool_true_false_are_defined) && (_MSC_VER < 1800)
 typedef int bool;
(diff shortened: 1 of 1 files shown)


ebassi:
The reason why I ignored it was because older g-ir-scanner versions would barf on `bool`.

I guess this means that graphene needs to depend on a minimum version of gobject-introspection.

lazka:
OK, CI uses Debian stretch and I don't see any errors/warnings there. So >= 1.50 should work for example.

ebassi:
Okay, let's merge this as is; I can add a requirement on 1.50 (and bump up the requirement for GLib 2.50) if that breaks elsewhere.

lazka:
Thanks!

PR #136, opened 2019-03-02
 matrix: Fix boolean logic in graphene_matrix_to_2d()

benjamin-otte:
This looks like a thinko, where && and || where confused leading to way
too many matrices being accepted as 2D when they shouldn't.

A new test, matrix/is-2d has been added.


benjamin-otte:
Updated the commit after finding `graphene_matrix_is_2d()` to use the same logic as that function, which was correct.

PR #139, opened 2019-03-05
Rect rounding improvements

3v1n0:
Proposed changes:

 - Add function for computing the area of a rect
 - Add function to compute the rounded extents
 - Fix docstring for round function
 - Add function to print rects

Test suite changes:
 - Test for `graphene_rect_round`
 - Test for `graphene_rect_get_area`
 - Test for `graphene_rect_round_extents`

Not sure if changing the round function would have been better, but I'd say no, as these two roundings cover different cases (for example, we'd need to round the extents when we want to pass a fractional rect for damaging the area).

### DIFF
--- a/src/graphene-rect.c
+++ b/src/graphene-rect.c
@@ -736,7 +736,8 @@ graphene_rect_round_to_pixel (graphene_rect_t *r)
  *
  * Rounds the origin and size of the given rectangle to
  * their nearest integer values; the rounding is guaranteed
- * to be large enough to contain the original rectangle.
+ * to be large enough to have an area bigger or equal to the
+ * original rectangle, but might not fully contain its extents.
  *
  * This function is the equivalent of calling `floor` on
  * the coordinates of the origin, and `ceil` on the size.
--- a/src/graphene-rect.c
+++ b/src/graphene-rect.c
@@ -434,6 +434,26 @@ GRAPHENE_RECT_GET (r, size, height)
 #undef GRAPHENE_RECT_GET
 
 /**
+ * graphene_rect_get_area:
+ * @r: a #graphene_rect_t
+ *
+ * Compute the area of given normalized rectangle.
+ *
+ * Returns: the area of the normalized rectangle
+ *
+ * Since: 1.10
+ */
+float
+graphene_rect_get_area (const graphene_rect_t *r)
+{
+  graphene_rect_t rr;
+
+  graphene_rect_normalize_r (r, &rr);
+
+  return rr.size.width * rr.size.height;
+}
+
+/**
  * graphene_rect_union:
  * @a: a #graphene_rect_t
  * @b: a #graphene_rect_t
--- a/src/graphene-rect.c
+++ b/src/graphene-rect.c
@@ -764,9 +764,9 @@ graphene_rect_round_to_pixel (graphene_rect_t *r)
  * This function is the equivalent of calling `floor` on
  * the coordinates of the origin, and `ceil` on the size.
  *
- * See also: graphene_rect_round_extents()
- *
  * Since: 1.4
+ *
+ * Deprecated: 1.10: Use graphene_rect_round_extents() instead
  */
 void
 graphene_rect_round (const graphene_rect_t *r,
@@ -808,8 +808,6 @@ graphene_rect_round (const graphene_rect_t *r,
  * size, then the move of the origin would not be compensated
  * by a move in the anti-origin, leaving the corners of the
  * original rectangle outside the rounded one.
- *
- * See also: graphene_rect_round()
  *
  * Since: 1.10
  */
--- a/src/graphene-rect.c
+++ b/src/graphene-rect.c
@@ -758,9 +758,13 @@ graphene_rect_round_to_pixel (graphene_rect_t *r)
  * their nearest integer values; the rounding is guaranteed
  * to be large enough to have an area bigger or equal to the
  * original rectangle, but might not fully contain its extents.
+ * Use graphene_rect_round_extents() in case you need to round
+ * to a rectangle that covers fully the original one.
  *
  * This function is the equivalent of calling `floor` on
  * the coordinates of the origin, and `ceil` on the size.
+ *
+ * See also: graphene_rect_round_extents()
  *
  * Since: 1.4
  */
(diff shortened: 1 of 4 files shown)


src/graphene-rect.c
```diff
@@ -433,6 +434,26 @@ GRAPHENE_RECT_GET (r, size, height)
...
 
+/**
+ * graphene_rect_get_area:
+ * @r: a #graphene_rect_t
+ *
+ * Compute the normalized area of given rectangle.
+ *
+ * Returns: the normalized area of the rectangle
+ *
+ * Since: 1.0
```
ebassi: `Since: 1.10`.

src/graphene-rect.c
```diff
@@ -433,6 +434,26 @@ GRAPHENE_RECT_GET (r, size, height)
 
 #undef GRAPHENE_RECT_GET
 
+/**
+ * graphene_rect_get_area:
+ * @r: a #graphene_rect_t
+ *
+ * Compute the normalized area of given rectangle.
```
ebassi: The rectangle is normalized; the area is not.

src/graphene-rect.c
```diff
@@ -756,6 +778,41 @@ graphene_rect_round (const graphene_rect_t *r,
...
+ * Rounds the origin of the given rectangle to its nearest
+ * integer value and and recompute the size so that the
+ * rectangle is large enough to contain all the conrners
+ * of the original rectangle.
+ *
+ * This function is the equivalent of calling `floor` on
+ * the coordinates of the origin, and recomputing the size
+ * calling `ceil` on the bottom-right coordinates.
+ *
+ * Since: 1.8
```
ebassi: `Since: 1.10`

src/graphene-rect.c
```diff
@@ -756,6 +778,41 @@ graphene_rect_round (const graphene_rect_t *r,
   res->size.height = ceilf (res->size.height);
 }
 
+/**
+ * graphene_rect_round_extents:
```
ebassi: I'm not entirely sure what's the purpose of this function, and how it relates to `graphene_rect_round()`. Care to explain a bit more? Maybe write an example in the documentation, so that it's clear which function should be used, and when.

src/graphene-rect.c
```diff
@@ -812,6 +869,25 @@ graphene_rect_interpolate (const graphene_rect_t *a,
   res->size.height = graphene_lerp (ra.size.height, rb.size.height, factor);
 }
 
+/**
+ * graphene_rect_print:
+ * @r: The rectangle to print
+ *
+ * Prints the coordinates and size of a rectangle.
+ *
+ * Since: 1.8
```
ebassi: `Since: 1.10`.

src/graphene-rect.h
```diff
@@ -123,6 +123,8 @@ GRAPHENE_AVAILABLE_IN_1_0
 float                   graphene_rect_get_width         (const graphene_rect_t *r);
 GRAPHENE_AVAILABLE_IN_1_0
 float                   graphene_rect_get_height        (const graphene_rect_t *r);
+GRAPHENE_AVAILABLE_IN_1_8
```
ebassi: `GRAPHENE_AVAILABLE_IN_1_10`.

src/graphene-rect.h
```diff
@@ -161,6 +163,9 @@ graphene_rect_t *       graphene_rect_round_to_pixel    (graphene_rect_t
 GRAPHENE_AVAILABLE_IN_1_4
 void                    graphene_rect_round             (const graphene_rect_t  *r,
                                                          graphene_rect_t        *res);
+GRAPHENE_AVAILABLE_IN_1_8
```
ebassi: `GRAPHENE_AVAILABLE_IN_1_10`.

src/graphene-rect.h
```diff
@@ -172,6 +177,9 @@ void                    graphene_rect_expand            (const graphene_rect_t
                                                          const graphene_point_t *p,
                                                          graphene_rect_t        *res);
 
+GRAPHENE_AVAILABLE_IN_1_8
```
ebassi: `GRAPHENE_AVAILABLE_IN_1_10`.

src/graphene-rect.c
3v1n0: So... As I wrote briefly in the PR description and hopefully a bit more in the commits, when you've cases in which you want to be sure that the rounded rectangle covers completely the area that was covered by the fractional rectangle (i.e. you want to damage that are) you need to make sure that you recompute the sizes taking in account the ceiled bottom-right angle coordinates.

Otherwise if the difference between the float coords and the floored coords is major than the difference between the ceiled size and and the fractional size, then we'd have a left/top shift which is not compensated by the size increase.

This happens also in the the round test, where the rectangle is properly rounded, and in fact its area is bigger than the original one, but the rounded one won't contain the original, because the left shift isn't covered enough by the size ceiling.

My initial thought was to just fix the `round` method, but since the APi was clearly done assuming only that we're flooring and ceilling, then I decided to add a different method for this.

So, said this, I'd be happy to be helped in wording this properly for the docs :)

3v1n0:
> You will need an initial commit to add the 1.10 versioning macros.

Yeah, it was also my guess, but I checked too late.
Fixed though :)

GeorgesStavracas:
I guess I can drop my very own commits to bump the version, then

src/tests/rect.c
```diff
@@ -181,6 +181,32 @@ GRAPHENE_TEST_UNIT_BEGIN (rect_inset)
...
+  graphene_rect_t r = GRAPHENE_RECT_INIT (0.5f, 1.9f,  9.3f, 8.7f);
+  graphene_rect_t e = GRAPHENE_RECT_INIT (0.0f, 1.0f, 10.0f, 9.0f);
+  graphene_rect_t rounded;
+
+  graphene_rect_round (&r, &rounded);
+  g_assert_cmpfloat (graphene_rect_get_area (&rounded),
+                     >=,
+                     graphene_rect_get_area (&r));
+  g_assert_false (graphene_rect_contains_rect (&rounded, &r));
+  g_assert_true (graphene_rect_equal (&rounded, &e));
```
ebassi: Please, use `graphene_assert_fuzzy_rect_equal()` with a `0.01f` epsilon.

src/graphene-version-macros.h
```diff
@@ -198,4 +199,19 @@
 # define GRAPHENE_AVAILABLE_IN_1_8              _GRAPHENE_PUBLIC
 #endif
 
+/* Graphene 1.10 */
+#if GRAPHENE_VERSION_MIN_REQUIRED >= GRAPHENE_VERSION_1_10
```
ebassi: These have been added already, so the commit is not necessary any more.

src/graphene-rect.h
```diff
@@ -172,6 +177,9 @@ void                    graphene_rect_expand            (const graphene_rect_t
                                                          const graphene_point_t *p,
                                                          graphene_rect_t        *res);
 
+GRAPHENE_AVAILABLE_IN_1_10
+void                    graphene_rect_print             (const graphene_rect_t *r);
```
ebassi: I'm not overly fond of these; I'm already regretting having added `graphene_matrix_print()`.

src/graphene-rect.c
ebassi: ⇑ What you wrote here should go in the description of the function. :smiley: 

> If you want to be sure that the rounded rectangle completely covers
> the area that was covered by the original rectangle—i.e. you want to
> cover the area including all its corners—this function will make sure
> that the size is recomputed taking into account the ceiling of the
> coordinates of the bottom-right corner. If the difference between the
> original coordinates and the coordinates of the rounded rectangle is
> greater than the difference between the original size and and the
> rounded size, then the move of the origin would not be compensated
> by a move in the anti-origin, leaving the corners of the original rectangle
> outside the rounded one.

Does that sound good?

I'd also honestly be happy with fixing the `round()` method, to be quite honest. Yes, it says what it does in the documentation, but it also says that the new rectangle would cover the old one, and that's clearly not always the case.

I'd also be happy with deprecating `round()` after adding `round_extents()`, if we want to play it extra-safe.

src/graphene-rect.h
3v1n0: So, want me drop this commit? :(

I mean, I understand it's just a dev-facility, but all the times I've to debug stuff I hate to loose precious seconds in doing all the `%d` repetitions around :-P

src/graphene-rect.c
3v1n0: As per the choice of deprecating `round()` or fixing it, I'm fine with the 2nd one.

But since you're the maintainer, and I don't want to risk introducing something that could break assumptions taken by other toolkits, I added a new one.

So let me know what you prefer. I can add the deprecation commit for now. If you then want to instead fix, I can take the other direction.

src/graphene-rect.h
ebassi: Yes, I really don't want this in the public API.

src/graphene-rect.c
ebassi: This looks good, thanks!

ebassi:
MSYS is apparently borked, so I'm going to ignore it for the time being, and merge this.

PR #140, opened 2019-03-06
Add Vala .vapi file generation

tintou:


ebassi:
Sorry, but this pull request is not going to be merged.

First of all, you should write a description for a merge request, summarising why do you think I should merge it in the first place; if you don't do this, I may legitimately come away with the impression that you don't value my time at all. This alone would be enough reason for me to close this MR.

Additionally, the CI breaks because—I assume, as I didn't really look at the MR—you did not do either of these two things:

 - update the CI environment to include Vala
 - add a configuration option to disable building the Vala API description on platforms that do not have Vala out of the box, like Windows or macOS

PRs with failing CI results are never going to be merged.

Finally, and the main reason this PR won't be merged: we are already shipping an API description in the form of introspection data. I'm not going to add another one that I don't use, I don't plan to use, and I don't want to maintain.

I'm afraid Vala will have to ship the VAPI file for Graphene alongside the rest of the VAPI files it already ships; or projects using Graphene with Vala will have to generate the VAPI from the introspection data themselves.

PR #141, opened 2019-03-07
rect: Add scale() method

GeorgesStavracas:
Scaling is a common operation when dealing with rectangles,
just as rounding.

Provide a scale() method to graphene_rect_t.

---

Proposed changes:

 - Inclusion of `graphene_rect_scale()`

Test suite changes:

 - Extend the test suite to stretch the new method


src/graphene-rect.c
```diff
@@ -829,3 +829,36 @@ graphene_rect_zero (void)
...
+ * @s_v: vertical scale factor
+ * @res: (out caller-allocates): return location for the
+ *   scaled rectangle
+ *
+ * Scales the size and origin of a rectangle horizontaly by @s_h,
+ * and vertically by @s_v. The result @res is normalized.
+ *
+ * Returns: (transfer none): a scaled rectangle
+ *
+ * Since: 1.8.6
```
ebassi: This is going to be `Since: 1.10` once I branch master.

src/graphene-rect.c
```diff
@@ -829,3 +829,36 @@ graphene_rect_zero (void)
...
+ */
+graphene_rect_t *
```
ebassi: This should be `void`.

src/graphene-rect.c
```diff
@@ -829,3 +829,36 @@ graphene_rect_zero (void)
...
+graphene_rect_scale (const graphene_rect_t *r,
+                     float                  s_h,
+                     float                  s_v,
+                     graphene_rect_t       *res)
+{
+  if (res != r)
+    *res = *r;
```
ebassi: You should use `graphene_rect_normalize_r (r, res)` and then modify `res`.

src/graphene-rect.h
```diff
@@ -175,6 +175,12 @@ void                    graphene_rect_expand            (const graphene_rect_t
 GRAPHENE_AVAILABLE_IN_1_4
 const graphene_rect_t * graphene_rect_zero              (void);
 
+GRAPHENE_AVAILABLE_IN_1_8
```
ebassi: This will need to be `GRAPHENE_AVAILABLE_IN_1_10`.

src/tests/rect.c
```diff
@@ -246,6 +246,49 @@ GRAPHENE_TEST_UNIT_BEGIN (rect_interpolate)
...
+    2.f, 5.f,
+    0.25f, 0.5f,
+    0.f, -10.f,
+  };
+
+  int i;
+
+  g_assert_true (G_N_ELEMENTS (unscaled) == G_N_ELEMENTS (scaled));
+
+  for (i = 0; i < G_N_ELEMENTS (unscaled); i++)
```
ebassi: You can use `for (int i = 0; …`. Graphene requires C99.

src/tests/rect.c
```diff
@@ -246,6 +246,49 @@ GRAPHENE_TEST_UNIT_BEGIN (rect_interpolate)
...
+  float scales[] = {
+    1.f, 1.f,
+    2.f, 5.f,
+    0.25f, 0.5f,
+    0.f, -10.f,
+  };
+
+  int i;
+
+  g_assert_true (G_N_ELEMENTS (unscaled) == G_N_ELEMENTS (scaled));
```
ebassi: You can use `g_assert_cmpint (G_N_ELEMENTS (unscaled), ==, G_N_ELEMENTS (scaled));` instead.

GeorgesStavracas: Done

src/tests/rect.c
GeorgesStavracas: Done

src/graphene-rect.c
GeorgesStavracas: Done

src/graphene-rect.c
GeorgesStavracas: I've added the commit bumping the version and adding the relevant macros. Let me know if that works for you.

src/graphene-rect.c
GeorgesStavracas: Done

src/graphene-rect.h
GeorgesStavracas: See https://github.com/ebassi/graphene/pull/141#discussion_r263202117

GeorgesStavracas:
Thanks for the quick review. This PR is now updated with the requested changes.

src/graphene-rect.c
```diff
@@ -829,3 +829,29 @@ graphene_rect_zero (void)
...
+                     float                  s_h,
+                     float                  s_v,
+                     graphene_rect_t       *res)
+{
+  graphene_rect_normalize_r (r, res);
+
+  res->origin.x *= s_h;
+  res->origin.y *= s_v;
+  res->size.width *= s_h;
+  res->size.height *= s_v;
```
ebassi: If the scaling factors are negative, the rectangle will not be normalized. I think you need to call normalize twice—or, at the very least, at the end of the scaling operation.

GeorgesStavracas: Hm, indeed. That should be correct now.

PR #146, opened 2019-04-15
build: use darwin_versions

tschoonj:
Fixes: separate issue not opened

Proposed changes:

 - `library()` should use `darwin_versions` to set the dylib versioning on macOS.

Benchmark results:

 - Before: no benchmarking was performed
 - After:

Test suite changes:

 - None


ebassi:
Thanks!

Issue #96, opened 2017-01-17
graphene_quaternion_to_angles: which convention is used?


sirop:
### Experienced behavior
```
quat = graphene_quaternion_init(quat, 0.7071, 0.7071, 0, 0 ); 
graphene_quaternion_to_angles (quat,  &yaw, &pitch, &chroll);
```
yields
```
yaw = 180 
pitch  = 0
roll = -90
```
### Expected behaviour
If compared against https://github.com/matthew-brett/transforms3d
```
In [48]: np.rad2deg(quat2euler([0,0.7071,0.7071,0],axes='rzyx'))
Out[48]: array([  90.,   -0.,  180.])

In [49]: np.rad2deg(quat2euler([0,0.7071,0.7071,0],axes='szyx'))
Out[49]: array([ -90.,    0., -180.])

In [50]: np.rad2deg(quat2euler([0,0.7071,0.7071,0],axes='rxyz'))
Out[50]: array([-180.,    0.,  -90.])

In [51]: np.rad2deg(quat2euler([0,0.7071,0.7071,0],axes='sxyz'))
Out[51]: array([ 180.,   -0.,   90.])

In [52]: np.rad2deg(quat2euler([0,0.7071,0.7071,0],axes='szxy'))
Out[52]: array([  90.,   -0.,  180.])

In [53]: np.rad2deg(quat2euler([0,0.7071,0.7071,0],axes='rzxy'))
Out[53]: array([ -90.,    0., -180.])

In [54]: np.rad2deg(quat2euler([0,0.7071,0.7071,0],axes='ryxz'))
Out[54]: array([ 180.,   -0.,   90.])

In [55]: np.rad2deg(quat2euler([0,0.7071,0.7071,0],axes='syxz'))
Out[55]: array([-180.,    0.,  -90.])

In [56]: np.rad2deg(quat2euler([0,0.7071,0.7071,0],axes='szxy'))
Out[56]: array([  90.,   -0.,  180.])
```
one finds no triple with one of the angles being -90 and the other 180 degrees.
`q[0]` corresponts to graphene's `q.w` .

I am pretty sure about https://github.com/matthew-brett/transforms3d as I checked `quat2euler(q,axes='rzyx')` case  before with quaternion and Yaw-Pitch-Roll output of my gyro.

### Operating system in use
debian 8

### SIMD implementation in use
SSE

sirop:
Just seen `graphene_euler_init_from_quaternion (...)` .
Will try this out.

[...]

PR #150, opened 2019-05-17
Add a test that roundtrips eulers via a matrix

alexlarsson:
This is currently failing, it seems like there is some kind of sign
issue in the matrix conversion.

I noted this in gthree, where its breaking the look_at function.


ebassi:
Yeah, it's likely connected to the fact that I implemented the quaternion/euler/matrix conversions wrong—see #96. I'll try to do an overhaul of the whole thing.

PR #157, opened 2019-06-19
graphene_quaternion_slerp: Always use the shortest path

alexlarsson:
As discussed in the "Invertin Quaternions" section in:
 http://www.euclideanspace.com/maths/algebra/realNormedAlgebra/quaternions/slerp/

We need to check if the dot product (i.e. cos(theta/2)) is
negative. If so, that means the rotation is > 180 degrees and
we should invert the rotation so as to take the shortest path.

Without this rotation animations in gthree get really weird when you
pass 360 degrees. I.e. interpolation is slow clockwise until 360
deg, where it switches to a very fast counter-clockwise rotation only
to then continue the slow clockwise rotation.


ebassi:
Yeah, this makes sense—also in light of 04b3aa3d48f7926ada3198e824a716ab2e086a8f.

ebassi:
The failure on Windows is unrelated.

PR #169, opened 2019-08-22
Fix graphene_ray_transform

alexlarsson:
We can't use graphene_matrix_transform_vec3() to transform the origin
because we need to point to respect transformations (but not the
directions, so we can use vec3 there instead of reconstructing
the direction).

Fixes #...

Proposed changes:

 - ...

Benchmark results:

 - Before: ...
 - After: ...

Test suite changes:

 - ...

### DIFF
--- a/src/graphene-matrix.c
+++ b/src/graphene-matrix.c
@@ -1090,13 +1090,13 @@ graphene_matrix_transform_ray (const graphene_matrix_t *m,
                                graphene_ray_t          *res)
 {
   graphene_vec3_t origin, direction;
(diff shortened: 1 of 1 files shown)


alexlarsson:
I added support for raycasting to gthree HEAD (see examples/interactive), and without this its totally broken. With this fix it mostl works although sometimes it seems something goes wrong and nothing is selected.

Also, part of the support was adding a bunch of ray-to-$primitive helpers, and these should probably be in graphene: https://github.com/alexlarsson/gthree/blob/master/gthree/gthreemesh.c#L10

PR #173, opened 2019-11-12
meson: Don't error with MSVC C4819 warning

pudding8757:
The compiler warning C4819 (encoding related warning) is a trivial
thing on non-English locale system. Instead of error out,
explicitly specify encoding as "utf-8".

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -79,7 +79,7 @@ if cc.get_id() == 'msvc'
     '-we4071',
     '-we4244',
     '-we4150',
-    '-we4819'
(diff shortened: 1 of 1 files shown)


nirbheek:
This was discovered in gstreamer: https://gitlab.freedesktop.org/gstreamer/cerbero/merge_requests/322

ebassi:
It does make sense to me. Thanks for the merge request, @pudding8757, and thanks to @nirbheek for the reference to the issue on GStreamer.

PR #174, opened 2019-12-01
Don't use obsolete isinff and isnanf functions

lantw44:
GNU libc already marks isinff and isnanf functions as obsolete. They
have been replaced by isinf and isnan macros defined by C99. Since we
already requires C99 in this project, it should be nice if we can
switch to standard macros.

Proposed changes:

 - Drop the use of non-standard and obsolete isinff and isnanf functions.

Benchmark results:

 - None

Test suite changes:

 - None

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -164,8 +164,6 @@ endif
 
 math_exts = [
   'sincosf',
-  'isinff',
-  'isnanf',
 ]
 
 foreach mfunc: math_exts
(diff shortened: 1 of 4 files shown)


lantw44:
It currently fails with GCC `__builtin_isinf` when -O1 or higher is used. I haven't found out why it fails. It works with Clang `__builtin_isinf` and FreeBSD libc. GNU libc just defines `isinf` as `__builtin_isinf`.

ebassi:
I initially used `isinf()` and `isnan()`, but those ended up breaking when using GCC because it tried to promote `float`s to `double`s, and the results wouldn't match any more—especially when using GCC vectorisation.

It could very well be a bug in GCC.

PR #177, opened 2020-01-07
meson: pick up gobject from a subproject fallback as well

tp-m:
Makes graphene find gobject in a gst-build on Windows scenario.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -215,7 +215,14 @@ build_gobject = false
 gobject_req_version = '>= 2.30.0'
 if get_option('gobject_types')
   graphene_gobject_api_path = '@0@-gobject-@1@'.format(meson.project_name(), graphene_api_version)
-  gobject = dependency('gobject-2.0', version: gobject_req_version, required: false)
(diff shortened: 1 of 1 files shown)


tp-m:
There's something else I need to investigate as well g-ir-scanner related

tp-m:
> There's something else I need to investigate as well g-ir-scanner related

Unrelated, this one should be good to go, I'll do a separate PR for other things.

ebassi:
I assume this works in the case where graphene is built as a sub-project of another Meson project that also depends on gobject?

It would be nice to leave a comment in the `meson.build` file to that effect.

tp-m:
> I assume this works in the case where graphene is built as a sub-project of another Meson project that also depends on gobject?

Correct. I'm trying to add graphene as subproject to GStreamers's gst-build (it's used in our OpenGL plugin in gst-plugins-base) and make things work on Windows where there's no system install of GObject.

> It would be nice to leave a comment in the `meson.build` file to that effect.

Done. (Plus some gratuitious line wrapping, since it was getting a bit long.)

PR #178, opened 2020-01-08
meson: don't build introspection if g-ir-scanner is not available

tp-m:


meson.build
```diff
@@ -226,7 +226,7 @@ endif
 
 # Optional dependency on GObject-Introspection; if GObject is disabled
 # then we don't build introspection data either
-build_gir = build_gobject and get_option('introspection') and get_option('default_library') == 'shared'
+build_gir = build_gobject and get_option('introspection') and get_option('default_library') == 'shared' and find_program('g-ir-scanner', required: false).found()
```
TingPing: Probably make it autodetecting feature if you are going to magically ignore what the builder configures.

tp-m: The current logic is also "autodetect", so as I see it this doesn't change the intention as it currently is.

I'd be happy to switch the `introspection` option (and others, if allowed) to a feature option, but I seem to remember that ebassi wasn't keen on feature options. I might be making this up though. It seems like a change with potentially more fallout anyway.

nirbheek: From what I remember, ebassi doesn't like _autodetection_ of features, but feature options don't always autodetect. F.ex., if the default value is `'enabled'` then there is no autodetection and the default behaviour is identical to a boolean option set to `true`, but with better internal behaviour (`required: get_option('introspection')`)

Even if the default value for your project's feature options is `'auto'`, you can pass `--auto-features=enabled` to force them all to be enabled and require explicit disabling of options. Distros already pass that option by default while building projects that use meson.

For users on non-Linux platforms, feature options set to `'auto'` are good for reducing developer friction, but that's entirely up to the maintainer. `'enabled'` is always better than `true`.

nirbheek: Another thing: there is a slight complication with changing options, I think it requires notification on some mailing list somewhere to the release team or something so that people maintaining build recipes (like the gnome runtime for flatpak) have notice? I don't remember the details.

ebassi: I'm okay with `feature` options, but the whole set of escape hatches is getting a bit ridiculous.

First of all, Meson already checks for `g-ir-scanner`, what's the point in projects using `gnome.gir()` checking again? Meson should disable the introspection generation if it can't find `g-ir-scanner` in the `PATH`. The `gnome` module in Meson should also take care of checking if the target library is a shared or a static object, and bail out if the latter.

Finally, if `introspection=true`, not finding the introspection scanner/compiler **should** be a build breaking event; in this case, making `introspection` a feature would at least allow us to bail out in the case it's explicitly enabled. The problem is, of course, that feature options don't accept `true`/`false`, so this will need changes in build recipes that explicitly enable/disable introspection.

ebassi:
@tp-m Any news on this?

tp-m:
> @tp-m Any news on this?

I'm not really sure what the outcome of the discussion was.

Should I change it to a feature option, even though Meson doesn't provide a migration path yet?

I kinda assumed this depends on https://github.com/mesonbuild/meson/issues/7444

ebassi:
> I'm not really sure what the outcome of the discussion was.

I'm okay with migrating to a feature option, but I'd like to wait on the Meson issue.

Additionally, I'd still like for Meson to automatically disable introspection if the tools are not available, instead of having to check myself.

ebassi:
The merge request in #200 has been merged, so I think we can close this.

Thanks!

PR #182, opened 2020-04-02
tests: installed tests should use TAP

rossburton:
The GNOME installed tests system expects the test output to be TAP:

"The key Output (if specified) can only take one value at present, which is TAP. This specifies that the test outputs Test Anything Protocol."
(https://wiki.gnome.org/Initiatives/GnomeGoals/InstalledTests)

Export MUTEST_OUTPUT=tap so this is the case.

Fixes #...

Proposed changes:

 - ...

Benchmark results:

 - Before: ...
 - After: ...

Test suite changes:

 - ...

### DIFF
--- a/tests/gen-installed-test.py
+++ b/tests/gen-installed-test.py
@@ -29,7 +29,11 @@ def write_template(filename, data):
         f.write(data)
 
 def build_template(testdir, testname):
-    return "[Test]\nType=session\nExec={}\n".format(os.path.join(testdir, testname))
+    return """[Test]
+Type=session
+Exec={}
+TestEnvironment=MUTEST_OUTPUT=tap;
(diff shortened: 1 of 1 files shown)


tests/gen-installed-test.py
```diff
@@ -29,7 +29,11 @@ def write_template(filename, data):
         f.write(data)
 
 def build_template(testdir, testname):
-    return "[Test]\nType=session\nExec={}\n".format(os.path.join(testdir, testname))
+    return """[Test]
+Type=session
+Exec={}
+TestEnvironment=MUTEST_OUTPUT=tap
```
ebassi, line 35: This is a string list, so it should be:

```suggestion
TestEnvironment=MUTEST_OUTPUT=tap;
```

PR #183, opened 2020-04-03
meson: Fix check for builtype arguments

nirbheek:
`get_option('buildtype')` will return `'custom'` for most combinations
of `-Doptimization` and `-Ddebug`, but those two will always be set
correctly if only `-Dbuildtype` is set. So we should look at those
options directly.

For the two-way mapping between `buildtype` and `optimization` + `debug`, see this table:
https://mesonbuild.com/Builtin-options.html#build-type-options

Same change as glib: https://gitlab.gnome.org/GNOME/glib/merge_requests/1433

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -188,11 +188,12 @@ endforeach
 
 # Debugging
 debug_flags = []
-buildtype = get_option('buildtype')
-if buildtype == 'release'
-  debug_flags += [ '-DG_DISABLE_ASSERT' ]
-elif buildtype.startswith('debug')
+if get_option('debug')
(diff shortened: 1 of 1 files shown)


nirbheek:
Test failure seems to be unrelated?

>  `4/21 graphene / box                          ERROR   0.988502025604248 s`

ebassi:
@nirbheek The transient failure in the `graphene_box_t` on MSYS2/x86 is known.

nirbheek:
Should I re-push till the test passes?

ebassi:
Nah, it's just annoying, and I care very little about 32bit platforms.

PR #186, opened 2020-06-15
matrix_init_look_at: Handle corner cases better

alexlarsson:
graphene_matrix_init_look_at() currently breaks badly in the case
where the given point is either the same as the eye, or in "up" direction
from the eye.

The exact definition of look_at() kinda breaks down in this situation,
but this is the behaviour three.js uses, and its better than nothing.


alexlarsson:
FTR, the green plane in examples/clipping in gthree master is fixed by this commit.

ebassi:
Nice; I do wonder if we have tests that have been commented out because of this.

PR #188, opened 2020-06-18
box: Use isinf() and signbit() when isinff() is not available

ebassi:
Instead of using bytewise comparison, use isinf() and hope it works
correctly with single precision floats on the target platform;
additionally, since the negative return value for negative infinities is
a GNU extension, use signbit() to check if the number is a positive or
negative infinity.

Proposed changes:

 - change the fallback path for the `is_empty()` and `is_infinity()` methods of `graphene_box_t`


src/graphene-box.c
```diff
@@ -436,18 +436,22 @@ graphene_box_is_empty (const graphene_box_t *box)
...
-   * bitwise comparison will not hold for infinities generated by
-   * other operations
-   */
-  int min_cmp = memcmp (&box->min.value, &pos_inf, sizeof (graphene_simd4f_t));
-  int max_cmp = memcmp (&box->max.value, &neg_inf, sizeof (graphene_simd4f_t));
-
-  return min_cmp == 0 && max_cmp == 0;
+  float vmin[3], vmax[3];
+
+  graphene_simd4f_dup_3f (box->min.value, vmin);
```
rmader, line 441: You could move these out of the `#ifdef` branch and save some duplicated lines (same below)

PR #189, opened 2020-06-18
Windows: use _isnanf macro instead of fallback

vtorri:


src/graphene-ray.c
```diff
@@ -52,6 +52,11 @@
 #include <math.h>
 #include <float.h>
 
+#if defined(_WIN32) && ! defined(isnanf)
```
nirbheek, line 55: Shouldn't this be `defined(_MSC_VER)` since I think MinGW does provide `isnanf`?

vtorri: mingw-w64 does not provide isnanf

```
$ fgrep -r isnanf /mingw64/x86_64-w64-mingw32/include/
/mingw64/x86_64-w64-mingw32/include/intrin.h:    __MACHINECE(int _isnanf(float))
/mingw64/x86_64-w64-mingw32/include/math.h:  extern int __cdecl __isnanf (float);
/mingw64/x86_64-w64-mingw32/include/math.h:  __CRT_INLINE int __cdecl __isnanf (float _x)
/mingw64/x86_64-w64-mingw32/include/math.h:        __isnanf(x),                                            \
/mingw64/x86_64-w64-mingw32/include/math.h:   _CRTIMP int __cdecl _isnanf(float _X);
```

so the check must not be changed

nirbheek: okay, thanks for checking :)

ebassi:
Sadly, this breaks on MSVC 32bit:

```
../src/graphene-ray.c(535): error C4013: '_isnanf' undefined; assuming extern returning int
```

vtorri:
accroding to msdn 

```The _isnan and _isnanf functions are Microsoft-specific. The _isnanf function is only available when compiled for x64.```

i'm going to fix this

vtorri:
arg ! i've written 32 bits instead of 64 bits

vtorri:
@ebassi do i close this PR and create another one (because of my wrong message) ?

ebassi:
> @ebassi do i close this PR and create another one (because of my wrong message) ?

No. You can edit the commit and then force push to your branch. The merge request will be updated.

ebassi:
I'd also squash both commits into a single one, so that we don't break the tests when bisecting.

PR #192, opened 2020-06-22
matrix: Clarify API docs of graphene_matrix_init_look_at()

alexlarsson:


alexlarsson:
http://www.songho.ca/opengl/gl_camera.html has some reference material on this, although the computed matrix there is the inverted final one.

alexlarsson:
Unclear why CI failed for a docs change, maybe just retry?

ebassi:
Thanks, looks good.

> Unclear why CI failed for a docs change, maybe just retry?

The 32bit MSYS2 build is flaky for unclear reasons. I just tend to ignore it.

PR #194, opened 2020-07-15
ARM NEON: Add support for Visual Studio ARM64 builds

fanc999:
Hi,

This attempts to add support for building Graphene with ARM NEON intrinsics for Visual Studio ARM64 builds, which is introduced with Visual Studio 2017 15.9.x.  This is done in a way like the previous work on the SSE intrinsics for Visual Studio, as there is no extension available that can alleviate us from using a `static inline` function when calling the ARM NEON intrinsics.

This involves:

*  Skip the checks for the presence of the ARM EABI macros and the ARM NEON macros, since those macros are done in a GCC-specific way and are not present (but it seems that Visual Studio offer the functionality that are demanded/offered by those macros).  The test code for ARM NEON intrinsics do compile as a result.  Check for `_M_ARM64` in graphene-config.h.meson as well when checking to see `GRAPHENE_HAS_ARM_NEON` is activated.

*  Add (Re-word) the intrinsics for Visual Studio ARM64 builds in `graphene-simd4f.h` and `graphene-simd4x4.h`, so that GCCisms do not get in the way.  This comes at the cost of using `static inline` functions where needed, as that would be the best we could do, as when we added the SSE intrinsics for Visual Studio builds.  Also add comments for the SSE intrinsics sections so that this distinction between compiler syntax differences is made clearer.

*  Fix building `graphene-ray.c`, where we define `isnanf()` to `_isnanf()` and activate `HAVE_ISNANF` on 64-bit Windows.  Change this to check for `_M_X64` (the Visual Studio macro defined for x64 builds) instead of `_WIN64`, since `_isnanf()` is provided on x64 builds only.

I was able to cross-compile Graphene for Windows 10 ARM with Visual Studio 2019 with ARM NEON intrinsics enabled, and have also built the code with Visual Studio 2019 on a Windows 10 ARM system and run the tests successfully, and the time taken seemed decent enough.

This does not attempt to cover introspection builds, as no official Python binaries exist for ARM64 Windows, so I was not able to build Graphene at this time with introspection support, but GLib/GObject are activated.

With blessings, thank you!

### DIFF
--- a/include/graphene-config.h.meson
+++ b/include/graphene-config.h.meson
@@ -18,7 +18,7 @@ extern "C" {
 #mesondefine GRAPHENE_HAS_SSE
 # endif
 
-# if defined(__ARM_NEON__)
+# if defined(__ARM_NEON__) || defined (_M_ARM64)
 #mesondefine GRAPHENE_HAS_ARM_NEON
 # endif
 
--- a/include/graphene-simd4f.h
+++ b/include/graphene-simd4f.h
@@ -468,7 +468,7 @@ typedef GRAPHENE_ALIGN16 union {
   }))
 
 /* On MSVC, we use static inlines */
-# elif defined (_MSC_VER)
+# elif defined (_MSC_VER) /* Visual Studio SSE intrinsics */
 
 /* Use static inline to inline all these functions */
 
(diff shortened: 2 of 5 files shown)


ebassi:
Thanks! This looks fantastic.

PR #195, opened 2020-07-24
Add CI using GitHub actions

ebassi:
Start with a CI pipeline using GitHub actions on a Ubuntu base.

### DIFF
--- a/.github/workflows/ubuntu-latest.yml
+++ b/.github/workflows/ubuntu-latest.yml
@@ -1,0 +1,23 @@
+name: Ubuntu/latest CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/ubuntu-latest.yml
```diff
@@ -0,0 +1,27 @@
...
+        sudo apt-get install gcc gir1.2-glib-2.0 gobject-introspection libgirepository1.0-dev libglib2.0-dev ninja-build python3-setuptools python3-wheel
+        sudo pip3 install meson==0.54.3
+    - name: configure the build
+      run: meson _build .
+    - name: build
+      run: ninja -C _build
+    - name: check
+      run: meson test -C _build
+    - name: dist
+      run: meson dist -C _build
```
nirbheek, line 27: To speed up the CI, if you require `meson==0.55.0` you can use `meson dist --no-tests` which won't build and test before making the dist.

ebassi: It's fast enough, for the moment, so I'm not really worried. But thanks for the heads up!

PR #198, opened 2020-08-13
Test meson fix for PYTHONIOENCODING

nirbheek:
Testing: https://github.com/mesonbuild/meson/pull/7589


nirbheek:
@ebassi seems to work fine without that meson PR, actually 🤔

PR #200, opened 2020-10-16
meson: Change introspection option to yielding feature

xclaesse:
This copies the same logic as in GTK+. A yielding feature option has the
advantage that if introspection is disabled in GTK and graphene is built
as subproject it will inherit the value from GTK main project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ using configuration options:
  * `-Darm_neon=false` - will disable the ARM NEON fast paths
  * `-Dgcc_vector=false` - will disable the GCC vector intrinsics
 
-If you don't plan on generating introspection data, use `-Dintrospection=false`
+If you don't plan on generating introspection data, use `-Dintrospection=disabled`
 when configuring Graphene; similarly, if you don't plan on using GObject with
 Graphene, use `-Dgobject_types=false`. Disabling GObject types will also
 automatically disable generating introspection data.
--- a/meson.build
+++ b/meson.build
@@ -234,7 +234,13 @@ endif
 
 # Optional dependency on GObject-Introspection; if GObject is disabled
 # then we don't build introspection data either
-build_gir = build_gobject and get_option('introspection') and get_option('default_library') == 'shared'
+gir = find_program('g-ir-scanner', required : get_option('introspection'))
+build_gir = build_gobject and gir.found()
+# Disable g-i when cross compiling by default because it rarely works unless
+# special care has been taken, in which case the user can enable the option.
+if meson.is_cross_build() and not get_option('introspection').enabled()
+  build_gir = false
+endif
 
 # Check for InitOnce on Windows
 if host_system == 'windows'
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -4,8 +4,9 @@ option('gtk_doc', type: 'boolean',
 option('gobject_types', type: 'boolean',
        value: true,
        description: 'Enable GObject types (depends on GObject)')
-option('introspection', type: 'boolean',
-       value: true,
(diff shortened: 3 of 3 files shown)


xclaesse:
The same logic got merged into gtk, pango, cairo, gstreamer, probably others.

xclaesse:
@ebassi CI failure seems unrelated.

meson.build
```diff
@@ -234,7 +234,8 @@ endif
 
 # Optional dependency on GObject-Introspection; if GObject is disabled
 # then we don't build introspection data either
-build_gir = build_gobject and get_option('introspection') and get_option('default_library') == 'shared'
+gir = find_program('g-ir-scanner', required : get_option('introspection'))
+build_gir = build_gobject and gir.found() and (not meson.is_cross_build() or get_option('introspection').enabled())
```
ebassi, line 238: This line is really hard to read. Additionally, `is_cross_build()` is not necessary, because anybody cross-compiling Graphene either explicitly disables introspection *or* have a native g-i build that they can use.

README.md
```diff
@@ -78,7 +78,7 @@ using configuration options:
  * `-Darm_neon=false` - will disable the ARM NEON fast paths
  * `-Dgcc_vector=false` - will disable the GCC vector intrinsics
 
-If you don't plan on generating introspection data, use `-Dintrospection=false`
+If you don't plan on generating introspection data, use `-Dintrospection=disabled`
 when configuring Graphene; similarly, if you don't plan on using GObject with
 Graphene, use `-Dgobject_types=false`. Disabling GObject types will also
```
ebassi, line 83: Maybe `gobject_types` should also be a feature option?

meson.build
xclaesse: The whole point here is to not have to explicitly disable when it's set to auto. That's the same line copied in many projects, I don't think graphene is different.

README.md
xclaesse: Maybe, but it's unrelated.

meson.build
ebassi: It's still an unreadable jumble of conditions that is hard to read.

xclaesse: IIRC native g-i is not enough since it compiles and run code which only works if you have set an exe wrapper in your cross file. With Meson >= 0.55 I think the condition `not meson.is_cross_build() or get_option('introspection').enabled()` could be replaced by `can_run_host_binaries()`.

matthiasclasen:
This would be nice to have for the GTK macos ci that I'm trying to set up

meson.build
ebassi: We can bump up the version of Meson to 0.55/0.56, if it means improving the legibility/correctness of this block. Especially if it avoids a new merge request (or, worse, downstream patch) from projects like Yocto to remove the `is_cross_build()` check.

In any case, and to make it even more painfully clear: my objection was that the line is hard to parse from a maintainer standpoint.

I don't immediately remember the Meson operator precedence, and I don't want to. Here we have four conditions, and one of them is negated? Or maybe both the ones in the parenthesis? The length of the line doesn't make it easier to read, either.

Can you please expand this to an `if...endif` block instead of golfing it? Something like:

```
build_gir = build_gobject
if build_gir
  if not get_option('introspection').enabled() or not gir.found()
    build_gir = false
  endif
  if not meson.can_run_host_binaries()
    build_gir = false
  endif
endif
```

xclaesse: Ok, I splitted the condition to make it more readable.

xclaesse: @ebassi does it look good to you now?

ebassi: Yes, this looks a lot better, thanks.

PR #202, opened 2020-12-10
Fix graphene_box_intersection for non-intersecting boxes.

### LINKED ISSUE #201: graphene_box_intersection returns incorrect results
I started trying to use `graphene_box_intersection` today, but found it returns true when it most definitely should return false.

This happens for non-intersecting boxes whose intersection returns a *negative* volume in one of the dimensions (`min.[xyz]` > `max.[xyz]`). So it's easy to work around and just ignore any intersection with higher min than max values. But of course, the function itself should ensure that never happens and returns false instead of true.

The source of the problem appears to be `graphene_box_intersection`'s use of:
```c
bool
(graphene_simd4f_cmp_ge) (const graphene_simd4f_t a,
                          const graphene_simd4f_t b)
{
  return a.x >= b.x &&
         a.y >= b.y &&
         a.z >= b.z &&
         a.w >= b.w;
}
```
Actually we need `a.x >= b.x || a.y >= b.y || a.z >= b.z` for correct operation of:
```c
bool
graphene_box_intersection (const graphene_box_t *a,
                           const graphene_box_t *b,
                           graphene_box_t       *res)
{
  graphene_simd4f_t min, max;

  min = graphene_simd4f_max (a->min.value, b->min.value);
  max = graphene_simd4f_min (a->max.value, b->max.value);

  if (graphene_simd4f_cmp_ge (min, max))
    {
      if (res != NULL)
        graphene_box_init_from_box (res, graphene_box_empty ());
      
      return false;
    }
  
  if (res != NULL)
    graphene_box_init_from_simd4f (res, min, max);
  
  return true;
}
```

vanvugt:
Fixes #201 

The old logic did essentially:
```
  (min.x >= max.x && min.y >= max.y && min.z >= max.z)
```
which for a pair of non-intersecting boxes that differ in only one or two
dimensions will return false when we need it to return true (meaning there
is no intersection). What we really need is:
```
  (min.x > max.x || min.y > max.y || min.z > max.z)
```
which can be rewritten as:
```
  !(min.x <= max.x && min.y <= max.y && min.z <= max.z)
```

### DIFF
--- a/src/graphene-box.c
+++ b/src/graphene-box.c
@@ -356,7 +356,7 @@ graphene_box_intersection (const graphene_box_t *a,
   min = graphene_simd4f_max (a->min.value, b->min.value);
   max = graphene_simd4f_min (a->max.value, b->max.value);
 
-  if (graphene_simd4f_cmp_ge (min, max))
+  if (!graphene_simd4f_cmp_le (min, max))
     {
       if (res != NULL)
         graphene_box_init_from_box (res, graphene_box_empty ());
--- a/tests/box.c
+++ b/tests/box.c
@@ -284,6 +284,9 @@ static void
 box_intersection (mutest_spec_t *spec)
 {
   graphene_box_t res;
+  graphene_box_t left, right;
+  graphene_box_t top, bottom;
+  graphene_box_t front, back;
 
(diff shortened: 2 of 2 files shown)


vanvugt:
The above failure does not appear to be related to the pull request.

ebassi:
Thanks!

PR #206, opened 2020-12-21
meson: only enable SSE2 on x86_64

xry111:
Now SSE2 is enabled for 32-bit x86 by default (with recent GCC).  It's
causing issues such as https://gitlab.gnome.org/GNOME/gtk/-/issues/3497.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -264,9 +264,9 @@ graphene_simd = []
 sse2_cflags = []
 if get_option('sse2')
   sse_prog = '''
-#if defined(__GNUC__) && (__GNUC__ < 4 || (__GNUC__ == 4 && __GNUC_MINOR__ < 9))
+#if defined(__GNUC__)
 # if !defined(__amd64__) && !defined(__x86_64__)
-#   error "Need GCC >= 4.9 for SSE2 intrinsics on x86"
+#   error "SSE2 intrinsics are only available on x86_64"
 # endif
 #elif defined (_MSC_VER) && !defined (_M_X64) && !defined (_M_AMD64)
 # error "SSE2 intrinsics not supported on x86 MSVC builds"


ebassi:
That's… A lot of changes, and not really what I'd like to merge.

meson_options.txt
```diff
@@ -10,8 +10,8 @@ option('introspection', type: 'boolean',
 option('gcc_vector', type: 'boolean',
        value: true,
        description: 'Enable GCC vector fast paths (requires GCC)')
-option('sse2', type: 'boolean',
-       value: true,
+option('sse2', type: 'feature',
```
ebassi, line 13: I don't want this to turn into a feature, sorry.

meson.build
```diff
@@ -262,7 +262,12 @@ graphene_simd = []
 
 # SSE intrinsics
 sse2_cflags = []
-if get_option('sse2')
+try_sse2 = get_option('sse2').enabled()
+if get_option('sse2').auto() and host_machine.cpu_family() == 'x86_64'
```
ebassi, line 266: I don't really want to have Meson test for this; the `sse_prog` below should be modified to fail is the architecture is not 64 bit.

README.md
```diff
@@ -75,7 +75,7 @@ Then run:
 It is possible, when building Graphene, to disable specific optimizations by
 using configuration options:
 
- * `-Dsse2=false` - will disable the SSE2 fast paths
+ * `-Dsse2=disable` - will disable the SSE2 fast paths
```
ebassi, line 78: Either all selections become features, or none do.

ebassi (changes requested):

meson_options.txt
xry111: So let's keep it a boolean.

My original thinking was to make it possible to enable SSE2 for 32-bit "forcefully" with `-Dsse2=enabled`.

meson.build
xry111: It's actually easier, just remove some conditions in `#if`.

meson_options.txt
ebassi: I'd rather not allow that; if we know that SSE types on 32bit architectures are problematic—especially when used by other libraries that don't have alignment guarantees—then I'd rather not give out footguns to people.

README.md
xry111: Let's keep them all booleans.

ebassi (approved):
Okay, this looks a lot better.

ebassi:
Thanks for the fixes!

PR #212, opened 2021-02-12
graphene-config: Hide SIMD implementation from GObject-Introspection

### LINKED ISSUE #211: GIR XML is architecture-dependent, perhaps unnecessarily?
### Experienced behavior

Graphene-1.0.gir conditionally defines `Graphene.HAS_GCC`, `Graphene.HAS_SCALAR`, `Graphene.HAS_SSE` constants from `graphene-config.h`.

On multiarch/multilib systems, this makes it one of the few GIR XML files that varies between architectures. Since GIR XML usually goes in /usr/share, this means it will either conflict between architectures, or give the wrong answer when asked about one of the architectures on a multiarch/multilib system.

(I know that the GIR XML for GLib also varies between architectures, but that one is easier to special-case since it comes from the gobject-introspection package.)

It looks as though `Graphene.SIMD_S` is also defined, and that one seems to be unconditionally `sse` even on non-x86, due to gobject-introspection's parser being fairly limited:

```
smcv@zelenka ~ % uname -a
Linux zelenka 4.19.0-14-s390x #1 SMP Debian 4.19.171-2 (2021-01-30) s390x GNU/Linux
smcv@zelenka ~ % python3
>>> from gi.repository import Graphene
>>> Graphene.SIMD_S
'sse'
```

### Expected behavior

These constants don't seem very useful to introspected code, so perhaps they should be omitted from the GIR XML.

### Steps to reproduce

* Build Graphene with GObject-Introspection on multiple architectures
* Compare the GIR XML with `diff`

### Operating system in use

Debian unstable (rolling release / Debian 11 alpha)

### SIMD implementation in use

Various - that's the problem :-)

smcv:
Exposing the various #mesondefine'd constants results in the GIR XML
unnecessarily varying between architectures.

Conversely, GObject-Introspection's rather simplistic cpp parser
thinks GRAPHENE_SIMD_S is always defined to "sse", even on
architectures where it should not be.

Introspected code is not going to be able to use SIMD anyway, so this
doesn't seem to have any value.

Resolves: https://github.com/ebassi/graphene/issues/211

---

Not tested yet, but hopefully fixes #211.

Proposed changes:

 - Wrap more of graphene-config.h in `#ifndef __GI_SCANNER__`

Benchmark results:

 - not benchmarked

Test suite changes:

 - Add a smoke-test for GObject-Introspection using PyGI

### DIFF
--- a/include/graphene-config.h.meson
+++ b/include/graphene-config.h.meson
@@ -11,41 +11,41 @@
 extern "C" {
 #endif
 
-#ifndef GRAPHENE_SIMD_BENCHMARK
+#ifndef __GI_SCANNER__
+# ifndef GRAPHENE_SIMD_BENCHMARK
 
-# if defined(__SSE__) || \
+#  if defined(__SSE__) || \
    (defined(_M_X64) && (_M_X64 > 0))
 #mesondefine GRAPHENE_HAS_SSE
 # endif
 
-# if defined(__ARM_NEON__) || defined (_M_ARM64)
+#  if defined(__ARM_NEON__) || defined (_M_ARM64)
 #mesondefine GRAPHENE_HAS_ARM_NEON
-# endif
+#  endif
 
-# if defined(__GNUC__) && (__GNUC__ >= 4 && __GNUC_MINOR__ >= 9) && !defined(__arm__)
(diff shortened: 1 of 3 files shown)


ebassi:
Looks good; thanks for adding the introspection test, it's much appreciated.

Issue #214, opened 2021-03-09
Graphene can't handle axis aligned ray directions


jadahl:
### Experienced behavior

Using `graphene_ray_intersect_box()` using a ray with a direction that is axis aligned (one or more components is `0.0`), the formula used to calculate whether the ray intersects with the box fails.

### Expected behavior

The ray/box intersection should be correctly detected.

### Steps to reproduce

Reproducer: 

```c
#include <graphene.h>
#include <stdio.h>

int
main (void)
{
  graphene_point3d_t min;
  graphene_point3d_t max;
  graphene_point3d_t origin;
  graphene_vec3_t direction;
  graphene_box_t box;
  graphene_ray_t ray;

  /* Box far to the top right */

  graphene_point3d_init (&min, 41.843132, 27.356903, -50.368336);
  graphene_point3d_init (&max, 51.698078, 29.080172, -50.368336);
  graphene_box_init (&box, &min, &max);

  /* Ray from (0, 0, 0) along the Y axis "upwards" *NOT* hitting the above box
   */

  graphene_point3d_init (&origin, 0, 0, 0);
  graphene_vec3_init (&direction, 0, 0.495176, -0.868793);
  graphene_ray_init (&ray, &origin, &direction);

  printf("1, intersects (exact): %d (should be 0)\n",
	 graphene_ray_intersects_box (&ray, &box));

  /* Nudged variant of the above ray */

  graphene_vec3_init (&direction, 0 + 0.0001, 0.495176, -0.868793);
  graphene_ray_init (&ray, &origin, &direction);
  printf("2, intersects (nudged): %d (should be 0)\n",
	 graphene_ray_intersects_box (&ray, &box));

  /* Box on the center top */

  graphene_point3d_init (&min, -5.654480, 27.356903, -50.368336);
  graphene_point3d_init (&max, 5.654475, 29.080172, -50.368336);
  graphene_box_init (&box, &min, &max);

  /* Ray from (0, 0, 0) along the Y axis "upwards" hitting the above box */

  graphene_point3d_init (&origin, 0, 0, 0);
  graphene_vec3_init (&direction, 0, 0.495176, -0.868793);
  graphene_ray_init (&ray, &origin, &direction);

  printf("3, intersects (exact): %d (should be 1)\n",
	 graphene_ray_intersects_box (&ray, &box));

  /* Nudged variant of the above ray */

  graphene_vec3_init (&direction, 2 * FLT_EPSILON, 0.495176, -0.868793);
[...]

PR #217, opened 2021-03-23
ray: Nudge ray direction of axis when intersecting

jadahl:
The formula used to calculate the inverse of the direction vector
doesn't handle the direction vector aligning with an axis. Depending on
the SIMD (or not SIMD) implementation used, a axis aligned vector would
either remain the same, or e.g. end up with NaN components messing up
any future calculations.

Fixing the math to handle this is non-trivial, so for now work around
this by nudging the direction vector slightly off axis so that it has a
better hand of hitting the right box even when the direction is axis
aligned.

Closes: #214

Fixes #...

Proposed changes:

 - ...

Benchmark results:

 - Before: ...
 - After: ...

Test suite changes:

 - ...

### DIFF
--- a/tests/ray.c
+++ b/tests/ray.c
@@ -167,6 +167,72 @@ ray_intersect_triangle (void)
 }
 
 static void
+ray_intersects_box (void)
+{
+  graphene_point3d_t min;
+  graphene_point3d_t max;
+  graphene_point3d_t origin;
+  graphene_vec3_t direction;
+  graphene_box_t box;
+  graphene_ray_t ray;
+
+  /* Off center box */
+
+  graphene_point3d_init (&min, 41.843132f, 27.356903f, -50.368336f);
+  graphene_point3d_init (&max, 51.698078f, 29.080172f, -50.368336f);
+  graphene_box_init (&box, &min, &max);
+
+  /* Ray from (0, 0, 0) along an axis *NOT* hitting the above box
+   */
+
+  graphene_point3d_init (&origin, 0, 0, 0);
+  graphene_vec3_init (&direction, 0, 0.495176f, -0.868793f);
+  graphene_ray_init (&ray, &origin, &direction);
+
+  mutest_expect ("intersection kind should be NONE",
+                 mutest_int_value (graphene_ray_intersects_box (&ray, &box)),
+                 mutest_to_be_false,
+                 NULL);
+
+  /* Nudged variant of the above ray */
+
+  graphene_vec3_init (&direction, 0 + 0.0001f, 0.495176f, -0.868793f);
+  graphene_ray_init (&ray, &origin, &direction);
+
+  mutest_expect ("intersection kind should still be NONE",
+                 mutest_int_value (graphene_ray_intersects_box (&ray, &box)),
+                 mutest_to_be_false,
+                 NULL);
+
+  /* Centered box */
+
+  graphene_point3d_init (&min, -5.654480f, 27.356903f, -50.368336f);
+  graphene_point3d_init (&max, 5.654475f, 29.080172f, -50.368336f);
+  graphene_box_init (&box, &min, &max);
+
+  /* Ray from (0, 0, 0) along the axis hitting the above box */
+
+  graphene_point3d_init (&origin, 0, 0, 0);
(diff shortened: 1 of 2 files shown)


tests/ray.c
```diff
@@ -166,6 +166,72 @@ ray_intersect_triangle (void)
...
+   */
+
+  graphene_point3d_init (&origin, 0, 0, 0);
+  graphene_vec3_init (&direction, 0, 0.495176f, -0.868793f);
+  graphene_ray_init (&ray, &origin, &direction);
+
+  mutest_expect ("intersection kind should be NONE",
+                 mutest_int_value (graphene_ray_intersects_box (&ray, &box)),
+                 mutest_to_be_false,
+		 NULL);
```
rmader, line 195: Indentation (also in the three cases below)

jadahl: Fixed.

tests/ray.c
```diff
@@ -166,6 +166,72 @@ ray_intersect_triangle (void)
...
+                 NULL);
+
+  /* Nudged variant of the above ray */
+
+  graphene_vec3_init (&direction, 0 + 0.0001f, 0.495176f, -0.868793f);
+  graphene_ray_init (&ray, &origin, &direction);
+
+  mutest_expect ("intersection kind should still be NONE",
+                 mutest_int_value (graphene_ray_intersects_box (&ray, &box)),
+                 mutest_to_be_false,
```
ebassi, line 204: This should be:

```suggestion
                 mutest_to_be, GRAPHENE_RAY_INTERSECTION_KIND_NONE,
```

tests/ray.c
```diff
@@ -166,6 +166,72 @@ ray_intersect_triangle (void)
...
+
+  /* Ray from (0, 0, 0) along the axis hitting the above box */
+
+  graphene_point3d_init (&origin, 0, 0, 0);
+  graphene_vec3_init (&direction, 0, 0.495176f, -0.868793f);
+  graphene_ray_init (&ray, &origin, &direction);
+
+  mutest_expect ("intersection kind should be ENTER",
+                 mutest_int_value (graphene_ray_intersects_box (&ray, &box)),
+                 mutest_to_be_true,
```
ebassi, line 221: This should be:

```suggestion
                 mutest_to_be, GRAPHENE_RAY_INTERSECTION_KIND_ENTER,
```

tests/ray.c
```diff
@@ -166,6 +166,72 @@ ray_intersect_triangle (void)
...
+  /* Ray from (0, 0, 0) along an axis *NOT* hitting the above box
+   */
+
+  graphene_point3d_init (&origin, 0, 0, 0);
+  graphene_vec3_init (&direction, 0, 0.495176f, -0.868793f);
+  graphene_ray_init (&ray, &origin, &direction);
+
+  mutest_expect ("intersection kind should be NONE",
+                 mutest_int_value (graphene_ray_intersects_box (&ray, &box)),
+                 mutest_to_be_false,
```
ebassi, line 194: This should be:

```suggestion
                 mutest_to_be, GRAPHENE_RAY_INTERSECTION_KIND_NONE,
```

tests/ray.c
```diff
@@ -166,6 +166,72 @@ ray_intersect_triangle (void)
...
+                 NULL);
+
+  /* Nudged variant of the above ray */
+
+  graphene_vec3_init (&direction, 2 * FLT_EPSILON, 0.495176f, -0.868793f);
+  graphene_ray_init (&ray, &origin, &direction);
+
+  mutest_expect ("intersection kind should still be ENTER",
+                 mutest_int_value (graphene_ray_intersects_box (&ray, &box)),
+                 mutest_to_be_true,
```
ebassi, line 231: This should be:

```suggestion
                 mutest_to_be_true, GRAPHENE_RAY_INTERSECTION_KIND_ENTER,
```

ebassi (changes requested):
The changes look good; a couple of fixes to the test suite to make it more idiomatic.

tests/ray.c
jadahl: The tested function is graphene_ray_intersect**s**_box(0 which returns a `bool`.

ebassi: You're right—damn my eyesight.

ebassi (approved):

PR #218, opened 2021-04-01
graphene-config: Fix GCC version check

doraskayo:
I expect the CI build to fail as it does for me locally. Creating as draft.


doraskayo:
I guess this will be covered by #219. Closing.

PR #219, opened 2021-04-01
Fixes for the GCC vector code

ebassi:
It seems it regressed even with the test suite running.


ebassi:
Merged as part of #220.

PR #220, opened 2021-04-02
simd4f: Make reciprocal operations 0-safe

doraskayo:
## Description
From the main commit messages: (performance measurements omitted)
```
    simd4f: Make reciprocal operations 0-safe
    
    Introduce a well-defined behavior for the reciprocal calculation of
    positive and negative 0. These would now result in positive and
    negative infinity, respectively.
    
    This behavior follows the definition of the IEEE 754 standard.
    
    In the SSE implementation, _mm_rcp_ps() already returned positive or
    negative infinity when used on a positive or negative 0. However, since
    it is then multiplied by 0, we end up with NaN. NaN is then used in the
    next calculations, which finally result in it being returned.
    
    To avoid ending up with NaN, we use a mask created by _mm_cmpeq_ps()
    and a constant, and along with _mm_andnot_ps() we replace infinity with
    0 in the multiplication. After the following subtraction, we end up
    with 2 multiplied by the positive or negative infinity, which finally
    returns the expected result.
    
    The GCC vectorization and scalar implementations are adjusted to return
    infinity when calculating the reciprocal of 0, with copysign() used to
    set the expected sign of the result.
    
    The ARM NEON implementation doesn't require adjusting as it is already
    0-safe.
    
    This comes at a cost of ~11.9% for the SSE implementation, while the
    performance of the GCC vectorization and scalar implementations
    largely remains the same.
```
```
    ray: Remove nudging workaround
    
    Since the reciprocals of positive and negative 0 are now well-defined,
    this workaround should no longer be needed.
```
### Tests
`ray-speed`: 50,000,000 ray intersection tests with a box, each using a unique ray.
`simd-speed`: 100,000,000 vector reciprocal calculations, each using a unique vector.
### Results
The following are measurements made for the entire PR. For measurements made for each individual commit, see the commit message.
```
SSE
  BEFORE
    1/2 graphene / ray-speed  OK             1.27s
    2/2 graphene / simd-speed OK             0.42s
  AFTER
    1/2 graphene / ray-speed  OK             1.12s
    2/2 graphene / simd-speed OK             0.47s

GCC VECTORIZATION
  BEFORE
    1/2 graphene / ray-speed  OK             1.32s
    2/2 graphene / simd-speed OK             1.12s
  AFTER
    1/2 graphene / ray-speed  OK             1.27s
    2/2 graphene / simd-speed OK             1.12s

SCALAR
  BEFORE
    1/2 graphene / ray-speed  OK             2.02s
    2/2 graphene / simd-speed OK             1.62s
  AFTER
    1/2 graphene / ray-speed  OK             1.87s
    2/2 graphene / simd-speed OK             1.32s
```
## Status
Ready. The following items were done:
- [x] Add unit tests.
- [x] Measure performance and optimize where possible.
- [x] Test the SSE implementation and adjust where necessary.
- [x] Test the GCC vectorization implementation and adjust where necessary.
- [x] Test the scalar implementation and adjust where necessary.
- [x] Test the ARM NEON implementation and adjust where necessary.

## Dependencies
The following PRs are included in this PR until merged:
- #219 for the GCC vectorization adjustments

### DIFF
--- a/include/graphene-simd4f.h
+++ b/include/graphene-simd4f.h
@@ -897,19 +897,19 @@ typedef int graphene_simd4i_t __attribute__((vector_size (16)));
 
 # define graphene_simd4f_cross3(a,b) \
   (__extension__ ({ \
-    const graphene_simd4f_t __a = (a); \
-    const graphene_simd4f_t __b = (b); \
-    graphene_simd4f_init (__a[1] * __b[2] - __a[2] * __b[1], \
-                          __a[2] * __b[0] - __a[0] * __b[2], \
-                          __a[0] * __b[1] - __a[1] * __b[0], \
+    const graphene_simd4f_t __cross_a = (a); \
+    const graphene_simd4f_t __cross_b = (b); \
+    graphene_simd4f_init (__cross_a[1] * __cross_b[2] - __cross_a[2] * __cross_b[1], \
+                          __cross_a[2] * __cross_b[0] - __cross_a[0] * __cross_b[2], \
+                          __cross_a[0] * __cross_b[1] - __cross_a[1] * __cross_b[0], \
                           0.f); \
   }))
 
 # define graphene_simd4f_dot3(a,b) \
   (__extension__ ({ \
-    const graphene_simd4f_t __a = (a); \
-    const graphene_simd4f_t __b = (b); \
-    const float __res = __a[0] * __b[0] + __a[1] * __b[1] + __a[2] * __b[2]; \
+    const graphene_simd4f_t __dot_a = (a); \
+    const graphene_simd4f_t __dot_b = (b); \
+    const float __res = __dot_a[0] * __dot_b[0] + __dot_a[1] * __dot_b[1] + __dot_a[2] * __dot_b[2]; \
     graphene_simd4f_init (__res, __res, __res, __res); \
   }))
 
--- a/include/graphene-simd4f.h
+++ b/include/graphene-simd4f.h
@@ -856,12 +856,15 @@ typedef int graphene_simd4i_t __attribute__((vector_size (16)));
 
 # define graphene_simd4f_reciprocal(v) \
   (__extension__ ({ \
+    _Pragma ("GCC diagnostic push") \
+    _Pragma ("GCC diagnostic ignored \"-Wfloat-equal\"") \
     (graphene_simd4f_t) { \
       (v)[0] != 0.f ? 1.f / (v)[0] : 0.f, \
       (v)[1] != 0.f ? 1.f / (v)[1] : 0.f, \
       (v)[2] != 0.f ? 1.f / (v)[2] : 0.f, \
       (v)[3] != 0.f ? 1.f / (v)[3] : 0.f, \
     }; \
+    _Pragma ("GCC diagnostic pop") \
   }))
 
 # define graphene_simd4f_sqrt(v) \
@@ -876,12 +879,15 @@ typedef int graphene_simd4i_t __attribute__((vector_size (16)));
 
 # define graphene_simd4f_rsqrt(v) \
   (__extension__ ({ \
+    _Pragma ("GCC diagnostic push") \
+    _Pragma ("GCC diagnostic ignored \"-Wfloat-equal\"") \
     (graphene_simd4f_t) { \
       (v)[0] != 0.f ? 1.f / sqrtf ((v)[0]) : 0.f, \
       (v)[1] != 0.f ? 1.f / sqrtf ((v)[1]) : 0.f, \
       (v)[2] != 0.f ? 1.f / sqrtf ((v)[2]) : 0.f, \
       (v)[3] != 0.f ? 1.f / sqrtf ((v)[3]) : 0.f, \
     }; \
+    _Pragma ("GCC diagnostic pop") \
   }))
 
 # define graphene_simd4f_add(a,b)       (__extension__ ({ (graphene_simd4f_t) ((a) + (b)); }))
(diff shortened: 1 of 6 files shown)


doraskayo:
CI seems to have issues setting up the environment due to `apt-get` failing to download `python3-lxml`:
> Err:16 http://security.ubuntu.com/ubuntu focal-updates/main amd64 python3-lxml amd64 4.5.0-1ubuntu0.2
  404  Not Found [IP: 52.154.174.208 80]
E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/l/lxml/python3-lxml_4.5.0-1ubuntu0.2_amd64.deb  404  Not Found [IP: 52.154.174.208 80]
Fetched 4249 kB in 0s (12.3 MB/s)
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
Error: Process completed with exit code 100.

doraskayo:
Unfortunately I don't own an ARM device, so I can't test the existing NEON implementation and check whether it needs to be adjusted as well. Looking at the pseudo-code in ARM's documentation and our existing NEON implementation of `graphene_simd4f_reciprocal`, I think it may already behave as expected without modification.

@ebassi, or someone else, I'd appreciate it if you could test this on an ARM device and confirm. I've added unit tests for the expected behavior under `tests/simd.c` for ease of testing.

doraskayo:
I've used a QEMU user emulated environment to test the NEON implementation, and it does seem to already be 0-safe without modification. However, it's better to confirm this on a bare-metal environment without a translation/emulation layer involved.

doraskayo:
I enabled NEON for AArch64 on Linux in #221, and included it in this PR for testing purposes. I want to avoid a situation where someone would test this branch on AArch64 hardware and confirm that the NEON implementation doesn't require adjustments without actually running the NEON code-path.

doraskayo:
I ended up installing Linux on an old `armv7l` phone with NEON support. As expected, the NEON implementation is already 0-safe and requires no additional modifications.

So it seems everything works :grinning:

doraskayo:
I removed the dependency on #221 as it was only added to ease the testing of the ARM NEON implementation. Now that it was fully tested, it's no longer needed.

src/graphene-simd4f.c
```diff
@@ -1178,10 +1180,10 @@ graphene_simd4f_t
...
 {
   graphene_simd4f_t s = {
-    fabsf (v.x) > FLT_EPSILON ? 1.0f / v.x : 0.f,
-    fabsf (v.y) > FLT_EPSILON ? 1.0f / v.y : 0.f,
-    fabsf (v.z) > FLT_EPSILON ? 1.0f / v.z : 0.f,
-    fabsf (v.w) > FLT_EPSILON ? 1.0f / v.w : 0.f
+    1.0f / v.x,
+    1.0f / v.y,
+    1.0f / v.z,
+    1.0f / v.w
```
doraskayo, line 1186: Following a comment from @jadahl, I looked into why this works a bit further and it seems that while IEEE 754 compliance is largely well-adopted by compilers in regards to floating-point division of negative and positive 0, it's still very much compiler-specific behavior as far as the C standard defines.

Specifically, Annex F of the C99 standard (ISO/IEC 9899:1999) defines floating-point division as conforming to the IEEE 754 standard (ISO/IEC 60559:1989). However, since Annex F is largely optional, compilers and C implementations are free to implement only some of it, or none of it.

So to be on the safe side, and to avoid breaking existing users with older or exotic compilers, I think I'll detect scenarios that aren't in conformance and add a fallback implementation which is more likely to be implemented as expected.

doraskayo: Done.

doraskayo:
I made a bunch of before/after performance measurements per commit (see commit messages) and for the entire PR (see PR description).

I think this is ready now. @ebassi, I'd appreciate your review and thoughts.

include/graphene-simd4f.h
```diff
@@ -854,15 +860,27 @@ typedef int graphene_simd4i_t __attribute__((vector_size (16)));
     (graphene_simd4f_t) { __val, __val, __val, __val }; \
   }))
 
+#if defined (HAVE_IEEE_754_FLOAT_DIVISION)
```
ebassi, line 863: This cannot work.

You're assuming Graphene's `config.h` is going to be included before `graphene-simd4f.h`, which is not going to be the case for projects that *use* Graphene.

You will need to do something like what we do for `GRAPHENE_HAS_SSE4`: the symbol must be defined in `graphene-config.h`; of course, since you determine the capability at compile time, this might not be doable.

You have three options:

 1. you check for the C standard version in use *and* the compiler in use
 2. you always use the fallback implementation
 3. you add a dependency on a `#define GRAPHENE_NO_FLOAT_DIV`, like we have a `GRAPHENE_SIMD_BENCHMARK` symbol that forces the SIMD implementation; projects that require some niche compiler can inject that symbol before including `graphene.h` and get the fallback implementation

doraskayo: Of course, you're right, it's a header. Unfortunately there's no specific macro or definition that can be relied upon in compile-time to detect this. I'll see what I can do.

doraskayo: @ebassi, since there was no apparent performance benefit from allowing division by zero in the GCC vectorization implementation, I went with (2) and dropped it.

I left the optimization only for the scalar implementation which is part of `libgraphene`'s compilation unit.

ebassi:
GitHub was having a fit every time I tried to run the CI, so I had to merge it manually.

Thanks for the changes!

PR #221, opened 2021-04-03
Enable NEON for AArch64 on Linux

doraskayo:
I actually don't have the hardware to test this, but in my QEMU user emulated environment it seems to allow AArch64 to use the NEON code-paths.

NEON for AArch64 has been supposedly enabled on Windows builds since https://github.com/ebassi/graphene/commit/29f9eb0245eac3e5c52ae9582984ae0f0d992553, but remained disabled on Linux builds.

### DIFF
--- a/include/graphene-config.h.meson
+++ b/include/graphene-config.h.meson
@@ -19,7 +19,7 @@ extern "C" {
 #mesondefine GRAPHENE_HAS_SSE
 # endif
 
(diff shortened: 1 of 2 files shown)


linkmauve:
This does enable NEON vector operations on my phone while in AArch64 mode, I never noticed those were missing before, thanks!

DjLizama (approved):

DjLizama (approved):

Issue #223, opened 2021-04-11
Wrong results for ray/box intersection on systems without isnanf


sbstnk:
### Experienced behavior
On systems without `isnanf`, such as those using musl instead of glibc, sometimes a ray is wrongly detected to intersect a box. This is breaking picking in gnome-shell on Alpine Linux or Void Linux (when using musl): https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/3976

This is due to a typo in the non-`HAVE_ISNANF` code path in `graphene_ray_intersect_box()`: https://github.com/ebassi/graphene/blob/a28aacb00a103ac616374c025406f629ed6a5289/src/graphene-ray.c#L566

This should be `ty_max < tx_max` as is already correctly done in the `HAVE_ISNANF` code path.

### Expected behavior
Only detect an intersection when there is one.

### Steps to reproduce
Try running the following program with glibc and with musl:

```
#include <graphene.h>

int main (int argc, char *argv[]) {
    graphene_point3d_t min, max;
    graphene_box_t box;

    graphene_point3d_init (&min, -14.001682, 50.620972, -50.368336);
    graphene_point3d_init (&max, 12.924316, 72.161774, -50.368336);
    graphene_box_init (&box, &min, &max);
   
    graphene_point3d_t o;
    graphene_vec3_t d;
    graphene_ray_t r;

    graphene_point3d_init (&o, 0.0, 0.0, 0.0);
    graphene_vec3_init (&d, -0.187414, 0.804487, -0.563628);

    graphene_ray_init (&r, &o, &d);

    if (graphene_ray_intersects_box (&r, &box)) {
        return 1;
    }

    return 0;
}
```

### Operating system in use
Fedora 34, but graphene compiled against musl instead of glibc to test this behavior.

### SIMD implementation in use
SSE2

PR #224, opened 2021-04-11
ray: Fix ray/box intersection on systems without isnanf

sbstnk:
Fixes https://github.com/ebassi/graphene/issues/223

Proposed changes:

 - Fix the typo in the non-`HAVE_ISNANF` code path to match the behavior of the `HAVE_ISNANF` code path.

### DIFF
--- a/src/graphene-ray.c
+++ b/src/graphene-ray.c
@@ -563,7 +563,7 @@ graphene_ray_intersect_box (const graphene_ray_t *r,
 #else
   if (ty_min > tx_min || fpclassify (tx_min) == FP_NAN)
     tx_min = ty_min;
-  if (ty_max > tx_max || fpclassify (tx_max) == FP_NAN)
+  if (ty_max < tx_max || fpclassify (tx_max) == FP_NAN)
     tx_max = ty_max;
 #endif
 
--- a/tests/ray.c
+++ b/tests/ray.c
@@ -232,6 +232,110 @@ ray_intersects_box (void)
                  NULL);
 }
 
+/* Adapted from Mutter/Clutter */
+static void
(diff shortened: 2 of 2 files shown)


Gottox:
It would be awesome to add a testcase for this issue.

ebassi:
Indeed, it would be awesome to have a test case. :-)

ericonr:
I think merging this as a simple patch is reasonable, but I would recommend going with https://github.com/ebassi/graphene/pull/225 later. I can rebase it.

sbstnk:
I wasn't sure about adding a test case, because due to the nature of the bug that would just be some *random* numbers for which it happens to fail instead of a real class of issues. But I can add the example from the bug report as a test. Can't think of a good name for such a test though.

sbstnk:
I was thinking of turning the code I used to find these random numbers into a test that does picking for ~1000 random coordinates like it would be done in clutter/mutter. Not sure how useful such non-deterministic tests would be though.

sbstnk:
I couldn't find a way to make a useful test with random positions, so I went with some hardcoded ones.

This includes code based on Mutter/Clutter which was written by @GeorgesStavracas. Since Mutter is GPL and graphene is MIT, are you okay with using that code under MIT @GeorgesStavracas?

GeorgesStavracas:
I'm fine with it :slightly_smiling_face:

ebassi:
I think both #225 and #220 have fixed most of these issues.

sbstnk:
Would it make sense to open a separate PR with just the test since a test was specifically requested here and the merged fix did not have one?

ebassi:
Yes, sure.

PR #225, opened 2021-04-11
ray: simplify nan checking by creating a macro.

### LINKED ISSUE #223: Wrong results for ray/box intersection on systems without isnanf

ericonr:
Avoids the #ifdef forest and code duplication resulting from it.
There was mismatch in the two code paths, see [1] and [2], and this
commit avoids repeating the same mistake.

[1] https://github.com/ebassi/graphene/issues/223
[2] https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/3976

Fixes #223

### DIFF
--- a/src/graphene-ray.c
+++ b/src/graphene-ray.c
@@ -52,11 +52,6 @@
 #include <math.h>
 #include <float.h>
 
-#if defined(_M_X64) && ! defined(isnanf)
-# define isnanf(x) _isnanf(x)
-# define HAVE_ISNANF
-#endif
-
 /**
  * graphene_ray_alloc: (constructor)
  *
@@ -555,17 +550,10 @@ graphene_ray_intersect_box (const graphene_ray_t *r,
   /* These lines also handle the case where tx_min or tx_max is NaN
    * (result of 0 * INFINITY): NaN != NaN
    */
-#ifdef HAVE_ISNANF
-  if (ty_min > tx_min || isnanf (tx_min))
+  if (ty_min > tx_min || graphene_isnan (tx_min))
     tx_min = ty_min;
-  if (ty_max < tx_max || isnanf (tx_max))
+  if (ty_max < tx_max || graphene_isnan (tx_max))
     tx_max = ty_max;
-#else
-  if (ty_min > tx_min || fpclassify (tx_min) == FP_NAN)
-    tx_min = ty_min;
-  if (ty_max > tx_max || fpclassify (tx_max) == FP_NAN)
-    tx_max = ty_max;
-#endif
 
   float tz_min, tz_max;
   if (graphene_vec3_get_z (&inv_dir) >= 0.f)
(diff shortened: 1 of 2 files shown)


Gottox:
Be aware of this comment: https://github.com/ebassi/graphene/pull/174#issuecomment-561325167

ericonr:
@Gottox I am still using `isnanf` when available, just avoiding the code duplication resulting from how it was originally implemented.

src/graphene-ray.c
```diff
@@ -53,8 +53,9 @@
 #include <float.h>
 
 #if defined(_M_X64) && ! defined(isnanf)
```
mixi, line 55: You are piggybacking a special case instead of handling it the same way as before. This goes against what you suggested in https://github.com/ebassi/graphene/pull/225#issuecomment-817358342.

This should be:
```
#if defined(_M_X64) && !defined(isnanf)
# define my_isnan(x) _isnanf(x)
#elif HAVE_ISNANF
# define my_isnan(x) isnanf(x)
#else
# define my_isnan(x) isnan(x)
#endif
```

ericonr: Thanks, I had only grepped for `HAVE_ISNANF`. I have now seen https://github.com/ebassi/graphene/blob/master/meson.build#L172

ericonr:
Ping?

src/graphene-ray.c
```diff
@@ -53,8 +53,11 @@
...
-# define isnanf(x) _isnanf(x)
-# define HAVE_ISNANF
+# define my_isnan(x) _isnanf(x)
```
ebassi, line 56: Would it be possible to hoist this whole thing into `graphene-private.h` and call it `graphene_isnan()` instead? Something like:

```c
#if defined(_M_X64) && !defined(isnanf)
# define graphene_isnan(x) _isnanf(x)
#elif defined(HAVE_ISNANF)
# define graphene_isnan(x) isnanf(x)
#else
# define graphene_isnan(x) isnan(x)
#endif
```

ebassi (changes requested):
Thanks for your contribution!

src/graphene-ray.c
ericonr: Definitely, I did it in the file directly because it was the only place using `isnan`.

Will let you know when I'm done.

ericonr: Done, thanks!

PR #229, opened 2021-06-05
build: Fix ARM64 (aarch64) builds on Visual Studio 2017

fanc999:
Hi,

From the commit message:

<i>On Visual Studio 2017, it is unfortunate that it does not allow us to include arm_neon.h directly for ARM64 builds, so we must include arm64_neon.h instead.</i>

<i>Interestingly, Visual Studio 2019 can include arm_neon.h directly for ARM64 builds, which will in turn include arm64_neon.h</i>

<i>Fix the Visual Studio 2017 ARM64 build by:</i>

*  <i>Updating the check program to include arm64_neon.h when building with Visual Studio 2017 for ARM64.</i>

*  <i>Use macros in graphene-config.h.meson so that we include the right header depending on the compiler being used when building for ARM64.</i>

With blessings, thank you!

### DIFF
--- a/include/graphene-config.h.meson
+++ b/include/graphene-config.h.meson
@@ -21,11 +21,6 @@ extern "C" {
 
 #  if defined(__ARM_NEON__) || defined (_M_ARM64) || defined (__aarch64__)
 #mesondefine GRAPHENE_HAS_ARM_NEON
-#   if defined (_MSC_VER) && (_MSC_VER < 1920) && defined (_M_ARM64)
-#    define GRAPHENE_ARM_NEON_HEADER <arm64_neon.h>
-#   else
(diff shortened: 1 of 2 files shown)


include/graphene-config.h.meson
```diff
@@ -21,6 +21,11 @@ extern "C" {
 
 #  if defined(__ARM_NEON__) || defined (_M_ARM64)
 #mesondefine GRAPHENE_HAS_ARM_NEON
+#   if defined (_MSC_VER) && (_MSC_VER < 1920) && defined (_M_ARM64)
```
ebassi, line 24: Since this header isn't used anywhere else, could you move this check into the block below?

fanc999: Moved.  I also removed the macro that is defined here and instead include the needed header directly according to the compiler and architecture used.

PR #230, opened 2021-06-06
Fix installation of the introspection.py file for installed-tests

mdevlamynck:
Fixes installed-test. The instrospection.py file was not installed which made the corresponding test fail to run.

### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -65,6 +65,8 @@ endif
 
 if build_gir and host_system == 'linux' and not meson.is_cross_build()
   foreach unit: ['introspection.py']
+    install_data(unit, install_dir: installed_test_bindir)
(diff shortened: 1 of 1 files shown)


tests/meson.build
```diff
@@ -65,6 +65,8 @@ endif
 
 if build_gir and host_system == 'linux' and not meson.is_cross_build()
   foreach unit: ['introspection.py']
+	install_data(unit, install_dir: installed_test_bindir)
```
ebassi, line 68: Could you fix the indentation to use two spaces?

mdevlamynck: Done, my bad.

ebassi: Thanks!

PR #236, opened 2021-09-05
Fix various broken macros when using GCC extension vectors

q66:
Commit 8e5c25109898fa4894df810a546b26c387eaae93 introduced some pragmas, however those unintentionally change the types of the macros, as the diagnostic pop pragma becomes the last statement and the type of the expression becomes void.


ebassi:
Thanks!

PR #241, opened 2021-12-15
Implement vector intrinsics for Clang

q66:
This implements vector intrinsics support to be used with Clang on platforms without SSE/NEON.


q66:
Do you think this might be worth combining under one set of macros? Doing so would have some advantages (a build made on GCC would be able to use intrinsics on Clang and the other way around)

q66:
Now combined for a common implementation. This should ensure that you can use different compilers and still have both use the intrinsics when using the library.

meson_options.txt
```diff
@@ -8,9 +8,9 @@ option('introspection', type: 'feature',
        value: 'auto',
        yield: true,
        description: 'Enable GObject Introspection (depends on GObject)')
-option('gcc_vector', type: 'boolean',
+option('vector_intrinsics', type: 'boolean',
```
ebassi, line 11: You should not rename the configuration option: this will break builds—just like it broke CI.

One option is to add a `vector_intrinsics` option and deprecate `gcc_vector`, but it requires bumping up the version of Meson to 0.60: https://mesonbuild.com/Build-options.html#deprecated-options

ebassi:
Thanks for looking into this; I'm in general okay with it, but the CI needs to pass first.

meson_options.txt
q66: sure, sounds good

q66:
I also wonder, what was the original reason they were disabled on 32-bit for GCC? Is this something that would also concern us on Clang? (I have no way to test any 32-bit targets right now)

ebassi:
Mainly, two reasons: compiler bugs with older versions of GCC, and basically no hardware to test outside of CI. To be fair, 32 bit architectures are essentially dead, and they aren't coming back, so I don't really care about them.

q66:
It is probably best to disable them universally anyway, since this would mean Clang-compiled builds enabling intrinsics could not be used on GCC

include/graphene-config.h.meson
```diff
@@ -36,9 +40,9 @@ extern "C" {
 # elif defined(GRAPHENE_HAS_ARM_NEON)
 #  define GRAPHENE_USE_ARM_NEON
 #  define GRAPHENE_SIMD_S "neon"
-# elif defined(GRAPHENE_HAS_GCC)
-#  define GRAPHENE_USE_GCC
```
ebassi, line 40: `GRAPHENE_USE_GCC` is an exported public symbol, so it cannot be removed.

You can define it alongside `GRAPHENE_USE_INTRINSICS`.

include/graphene-config.h.meson
```diff
@@ -24,7 +24,11 @@ extern "C" {
 #  endif
 
 #  if defined(__GNUC__) && (__GNUC__ >= 5 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 9)) && !defined(__arm__)
-#mesondefine GRAPHENE_HAS_GCC
+#mesondefine GRAPHENE_HAS_INTRINSICS
+#  endif
+
+#  if defined(__clang__)
+#mesondefine GRAPHENE_HAS_INTRINSICS
```
ebassi, line 31: `GRAPHENE_HAS_GCC` is an exported public symbol, so it cannot be removed.

You should still define it alongside `GRAPHENE_HAS_INTRINSICS` for the GCC branch.

ebassi (changes requested):

q66:
Alright, this should work I think.

include/graphene-config.h.meson
q66: alright

q66:
Now no symbols are being dropped

ebassi (approved):

PR #242, opened 2022-01-06
build-sys: do not cc.run() on cross-compiling

elmarco:
Enable IEEE 754 with clang & gcc without running test. We may want to
add an extra option if necessary.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -272,8 +272,14 @@ int main() {
  return 0;
 }
 '''
+if meson.is_cross_build()
+  ieee754_float_div = meson.get_external_property('ieee754_float_div', cc.get_id() in ['gcc', 'clang'])
+  message('Cross-building, assuming IEEE 754 division:', ieee754_float_div)
+else
+  ieee754_float_div = cc.run(float_division_prog, name: 'IEEE 754 floating-point division').returncode() == 0
+endif
 conf.set('HAVE_IEEE_754_FLOAT_DIVISION',
-  cc.run(float_division_prog, name: 'IEEE 754 floating-point division').returncode() == 0,
+  ieee754_float_div,
   description: 'Define if floating-point division is compliant with IEEE 754',
 )
 


elmarco:
fwiw this prevents cross-building gtk on fedora 35 atm,
https://gitlab.gnome.org/malureau/gtk/-/jobs/1714142

(fedora rawhide has mingw-graphene 1.10.6 without this issue)

meson.build
```diff
@@ -272,8 +272,14 @@ int main() {
  return 0;
 }
 '''
+if meson.is_cross_build()
+  have_ieee754 = cc.get_id() in ['gcc', 'clang']
```
ebassi, line 276: I think you want [`meson.get_cross_property()`](https://mesonbuild.com/Reference-manual_builtin_meson.html#mesonget_cross_property) here:

```
ieee754_float_div = meson.get_cross_property('ieee754_float_div', cc.get_id() in ['gcc', 'clang'])
```

and use `ieee754_float_div` instead of `have_ieee754`.

elmarco: good idea

elmarco: I wonder if it fits there, as you may break it with extra cflags I guess

ebassi: If you pass additional compiler flags without checking the implications then there's not much we can do about it

meson.build
```diff
@@ -272,8 +272,14 @@ int main() {
  return 0;
 }
 '''
+if meson.is_cross_build()
+  ieee754_float_div = meson.get_external_property('ieee754_float_div', cc.get_id() in ['gcc', 'clang'])
+  warning('Cross-building, assuming IEEE 754 division:', ieee754_float_div)
```
ebassi, line 277: I don't think we want to warn, here. If you're building with Meson's warnings turned to errors—e.g. in CI—you're going to break perfectly valid builds.

This should be a message, like:

```suggestion
  message('IEEE 754 floating-point division (cross build):', ieee754_float_div)
```
```

elmarco:
Yeah, I was thinking the same. But I am less familiar with github CI. Let me check.

ebassi:
> But I am less familiar with github CI. Let me check.

Thanks, and no worries; I can also have a look.

Let's get this in, in any case.

meson.build
eli-schwartz: Or even summary().

DjLizama (approved):

PR #254, opened 2022-09-16
Add support for using g_autoptr with graphene types

tintou:
Proposed changes:

 - add g_autoptr(graphene_$TYPE_t) support

This is handy when writing an application.

### DIFF
--- a/include/graphene-gobject.h
+++ b/include/graphene-gobject.h
@@ -35,84 +35,118 @@ G_BEGIN_DECLS
 GRAPHENE_AVAILABLE_IN_1_0
 GType graphene_point_get_type (void);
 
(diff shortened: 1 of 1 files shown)


ebassi:
To be honest, if you're writing a C application I'd recommend to use Graphene types on the stack instead of using the heap; using `g_autoptr` works only if you're copying data from out of properties or signals.

tintou:
Yes.

In my use-case I have a GList of points so I wanted to use a `g_autolist(graphene_point_t) my_list`

ebassi:
I'm sorry to say, but a linked list of points is probably the most inefficient way you could find. I'd recommend a GArray, if you wanted to have the equivalent of a resizing C array of `graphene_point_t`.

Having said that, I don't have any particular objections to this pull request.

PR #255, opened 2022-12-18
meson: Override dependencies to improve usage as a subproject

nirbheek:
With this change, graphene can be consumed as a subproject without
making any changes to the build files of a project. All you need to do
is provide a wrap file with a `[provide]` section:

https://mesonbuild.com/Wrap-dependency-system-manual.html#provide-section

This is also necessary because otherwise projects need to hard-code
the subproject name, which might be `graphene` when using `wrap-git` or
`graphene-1.10.8` when using `wrap-file` (to build from a release
tarball). This can cause conflicts between different subprojects that
consume graphene differently.

Other projects like glib, cairo, pango, etc already do this.

Also, stop using fallback: kwarg for deps that don't need it.


ebassi:
Thanks!

Issue #204, opened 2020-12-15
2d bounding boxes


matthiasclasen:
GTK uses graphene_rect_t for 2d bounding boxes in a number of places, but its representation as origin+size makes is somewhat unsuitable for that use. Issues:
- due to floating point rounding, a a rect around p1, p2 may not actually contain p2
- intersecting rects with empty interior (such as bounding boxes for axis-aligned line segments) returns FALSE

graphene_box_t seems to be made for bounding box uses, and uses a min-max representation internally, but it has that extra z. It would be nice to have a graphene_box2d_t

matthiasclasen:
I have no deep thoughts on that. The main operations I foresee on a bounding box type are:
- containment/overlap checks (for points and rects)
- unions

ebassi:
The PR in #259 contains an implementation of an AABB.

PR #259, opened 2023-08-10
Add 2D axis-aligned bounding box type

ebassi:
Works like `graphene_box_t`, but in two dimensions.

Fixes: #204


include/graphene-box2d.h
```diff
@@ -0,0 +1,122 @@
...
+ * A 2D box, described as the axis-aligned area between a minimum and
+ * a maximum vertices lying on the same plane.
+ *
+ * Since: 1.12
+ */
+struct _graphene_box2d_t
+{
+  /*< private >*/
+  GRAPHENE_PRIVATE_FIELD (graphene_vec2_t, min);
+  GRAPHENE_PRIVATE_FIELD (graphene_vec2_t, max);
```
benjamin-otte, line 30: Should this be a single `graphene_vec4_t` because the vec2's are just vec4's in disguise afaik?

[The Rect type in GSK's Vulkan shader](https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gsk/vulkan/resources/rect.glsl) uses a vec4 itself, and that worked really well.

include/graphene-box2d.h
```diff
@@ -0,0 +1,122 @@
...
+                                                                     graphene_point_t       *center);
+GRAPHENE_AVAILABLE_IN_1_12
+void                      graphene_box2d_get_min                    (const graphene_box2d_t *box,
+                                                                     graphene_point_t       *min);
+GRAPHENE_AVAILABLE_IN_1_12
+void                      graphene_box2d_get_max                    (const graphene_box2d_t *box,
+                                                                     graphene_point_t       *max);
+GRAPHENE_AVAILABLE_IN_1_12
+void                      graphene_box2d_get_vertices               (const graphene_box2d_t *box,
+                                                                     graphene_vec2_t         vertices[]);
```
benjamin-otte, line 97: A `graphene_box2d_to_float (box, float[4])` would come in handy here for uploading to GLSL.

include/graphene-box2d.h
```diff
@@ -0,0 +1,122 @@
...
+                                                                     const graphene_point_t *point,
+                                                                     graphene_box2d_t       *res);
+GRAPHENE_AVAILABLE_IN_1_12
+void                      graphene_box2d_expand_vec2                (const graphene_box2d_t *box,
+                                                                     const graphene_vec2_t  *vec,
+                                                                     graphene_box2d_t       *res);
+GRAPHENE_AVAILABLE_IN_1_12
+void                      graphene_box2d_expand_scalar              (const graphene_box2d_t *box,
+                                                                     float                   scalar,
+                                                                     graphene_box2d_t       *res);
```
benjamin-otte, line 69: a linear transform function / fused multiply-add ala `scale * box + offset` is the one thing I've needed in the GSK clip rects to handle affine transforms. And it works as a fallback for all the other transforms (scales, offsets, etc), too.

include/graphene-box2d.h
```diff
@@ -0,0 +1,122 @@
...
+GRAPHENE_AVAILABLE_IN_1_12
+void                      graphene_box2d_union                      (const graphene_box2d_t *a,
+                                                                     const graphene_box2d_t *b,
+                                                                     graphene_box2d_t       *res);
+GRAPHENE_AVAILABLE_IN_1_12
+bool                      graphene_box2d_intersection               (const graphene_box2d_t *a,
+                                                                     const graphene_box2d_t *b,
+                                                                     graphene_box2d_t       *res);
```
benjamin-otte, line 77: intersection checking is the one absolutely most critical operation we are doing when rendering in GTK.

In fact, in certain workloads (think a GtkListbox in a scrolledwindow, like `gtk4-demo --run=listbox`) intersecting alone can take >50% of time spend in `gsk_renderer_render()` (according to sysprof).

So having an (inline?) function that just does an intersection check and returns `true/false` would be very neat. Something like `bool graphene_box2d_intersects (a, b)`

include/graphene-box2d.h
```diff
@@ -0,0 +1,122 @@
...
+graphene_box2d_t *        graphene_box2d_init_from_vectors          (graphene_box2d_t       *box,
+                                                                     unsigned int            n_vectors,
+                                                                     const graphene_vec2_t  *vectors);
+GRAPHENE_AVAILABLE_IN_1_12
+graphene_box2d_t *        graphene_box2d_init_from_box              (graphene_box2d_t       *box,
+                                                                     const graphene_box2d_t *src);
+GRAPHENE_AVAILABLE_IN_1_12
+graphene_box2d_t *        graphene_box2d_init_from_vec2             (graphene_box2d_t       *box,
+                                                                     const graphene_vec2_t  *min,
+                                                                     const graphene_vec2_t  *max);
```
benjamin-otte, line 56: Transitioning GTK code from using `graphene_rect_t` would benefit from `graphene_box2d_init_from_rect()` and `graphene_rect_init_from_box2d()`.

include/graphene-box2d.h
ebassi: Partial lane reads are always slower than full lane ones; if we store the min/max in the same SIMD type like this:

```
| min_x | min_y | max_x | max_y |
```

then it should be possible to optimise the access using:

```c
graphene_simd4f_t min = graphene_simd4_merge_low (box->value, graphene_sim4f_zero ());
graphene_simd4f_t max = graphene_simd4f_merge_high (box->value, graphene_simd4f_zero ());
```

include/graphene-box2d.h
```diff
@@ -0,0 +1,122 @@
...
+                                                                     graphene_vec2_t        *size);
+GRAPHENE_AVAILABLE_IN_1_12
+void                      graphene_box2d_get_center                 (const graphene_box2d_t *box,
+                                                                     graphene_point_t       *center);
+GRAPHENE_AVAILABLE_IN_1_12
+void                      graphene_box2d_get_min                    (const graphene_box2d_t *box,
+                                                                     graphene_point_t       *min);
+GRAPHENE_AVAILABLE_IN_1_12
+void                      graphene_box2d_get_max                    (const graphene_box2d_t *box,
+                                                                     graphene_point_t       *max);
```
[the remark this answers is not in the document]
ebassi: A fast path copy operation for both vertices would probably be okay, in the same way as `sincos()`; I mainly kept the same API as `graphene_box_t`, but I can see it being useful there, too.

include/graphene-box2d.h
ebassi: I started adding a static inline `intersects()` operator, but accessing private fields in the header file gets gnarly. I'm open to that, but it needs some more finesse than just giving access to the contents of the structure to everyone.

include/graphene-box2d.h
ebassi: What's the order of operations? Scale first, then offset? Or offset first, and then scale?

include/graphene-box2d.h
benjamin-otte: With `graphene_rect_t` we have basically inlined everything, see https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gsk/gskrectprivate.h for what we do there.

And it's a massive win for performance, because the compiler can not just avoid function calls, but it can inline the entire code, which means it can reorder instructions any way it likes.  

We measured significant speedups from that in the critical path (text rendering), which runs millions of times per second when benchmarking and simple graphene function calls like `graphene_rect_scale()` took multiple percentage points of total performance according to sysprof.

Here's roughly that example with GTK git main:

```
diff --git a/gsk/gpu/gskgpunodeprocessor.c b/gsk/gpu/gskgpunodeprocessor.c
index 7467ee72e1..2cb70d8a7c 100644
--- a/gsk/gpu/gskgpunodeprocessor.c
+++ b/gsk/gpu/gskgpunodeprocessor.c
@@ -3104,14 +3104,18 @@ gsk_gpu_node_processor_add_glyph_node (GskGpuNodeProcessor *self,
                                                  &glyph_bounds,
                                                  &glyph_offset);
 
-      glyph_tex_rect = GRAPHENE_RECT_INIT (-glyph_bounds.origin.x / scale,
-                                           -glyph_bounds.origin.y / scale,
-                                           gsk_gpu_image_get_width (image) / scale,
-                                           gsk_gpu_image_get_height (image) / scale);
-      glyph_bounds = GRAPHENE_RECT_INIT (0,
-                                         0,
-                                         glyph_bounds.size.width / scale,
-                                         glyph_bounds.size.height / scale);
+      graphene_rect_scale (&GRAPHENE_RECT_INIT (-glyph_bounds.origin.x,
+                                                -glyph_bounds.origin.y,
+                                                gsk_gpu_image_get_width (image),
+                                                gsk_gpu_image_get_height (image)),
+                           1 / scale, 1 / scale,
+                           &glyph_tex_rect);
+      graphene_rect_scale (&GRAPHENE_RECT_INIT (0,
+                                                0,
+                                                glyph_bounds.size.width,
+                                                glyph_bounds.size.height),
+                           1 / scale, 1 / scale,
+                           &glyph_bounds);
       glyph_origin = GRAPHENE_POINT_INIT (glyph_origin.x - glyph_offset.x / scale,
                                           glyph_origin.y - glyph_offset.y / scale);
 ```

With that patch applied, `GDK_DEBUG=no-vsync GSK_DEBUG=full-redraw build/demos/widget-factory/gtk4-widget-factory` of a `-O3` GTK build drops from ~1850fps to ~1780fps - which is somewhere around 3-5%. And here's what sysprof says:

image

Comment (ebassi) [reply include/graphene-box2d.h:81]: I'm not against inlining things—after all, that's why all `graphene_simd4f_t` functions are inline instead of using the pixman approach of a run time check for the CPU architecture. The main issue is that the higher order vectorised types in Graphene have private fields from the perspective of the caller, which makes writing static inlines harder. Unlike `graphene_rect_t`, which has public fields for a reason, other types like `graphene_vec[234]_t` and `graphene_box2d_t` have their public fields mangled when not compiling Graphene.

In any case, I've added a `graphene_box2d_intersects()` which inlines *most* of the intersection; it still has to call `graphene_box2d_get_minmax()` to get the minimum and maximum vertices; shaving more calls and getting direct access to the SIMD type gets complicated unless I completely abandon the field name mangling.

File: include/graphene-box2d.h (hunk 3 above)
Comment (ebassi) [reply include/graphene-box2d.h:73]: Another question: should the scale be a single factor, or should be it be decomposed into a vec2?

Basically something like this:

```c
void
graphene_box2d_scale_offset (const graphene_box2d_t *box,
                             const graphene_vec2_t *scale,
                             const graphene_point_t *offset,
                             graphene_box2d_t *res);
```

or with a single `float scale` to be applied to both coordinates?

benjamin-otte: Yeah, that would work. GTK uses 2 different floats for the scale factor everywhere.

(Ideally there'd be a `GRAPHENE_VEC2_INIT_SAME()` macro to easily do the 1 float => 2 floats conversion, but I suppose that's a discussion for a different issue.)

PR #260, opened 2023-08-19
build: Handle whitespaces in paths

sidt4:
Build fails due to this issue if source path contains spaces.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -101,7 +101,7 @@ if build_gir
   identfilter_py = join_paths(meson.current_source_dir(), 'identfilter.py')
 
   gir_extra_args = [
-    '--identifier-filter-cmd=@0@ @1@'.format(python.full_path(), identfilter_py),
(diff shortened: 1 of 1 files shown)


src/meson.build
```diff
@@ -98,7 +98,7 @@ graphene_dep_sources = []
 # Introspection
 if build_gir
   python = python.find_installation('python3')
-  identfilter_py = join_paths(meson.current_source_dir(), 'identfilter.py')
+  identfilter_py = '"' + join_paths(meson.current_source_dir(), 'identfilter.py') + '"'
 
   gir_extra_args = [
     '--identifier-filter-cmd=@0@ @1@'.format(python.full_path(), identfilter_py),
```
ebassi, line 104: Does it work if you put the quotes here, instead:

```suggestion
    '--identifier-filter-cmd="@0@ @1@"'.format(python.full_path(), identfilter_py),
```

sidt4: `"@0@ @1@"` doesn't work, but `@0@ "@1@"` does. I've updated the patch.

PR #266, opened 2024-04-19
fix incorrect use of python package

### LINKED ISSUE #265: Windows build fails when adding introspection integration
I have encountered an issue on Windows with `MSVC` and `-Dintrospection=enabled`.
The build fails on:
`src\meson.build:100:2: ERROR: <ExternalProgram 'python3' -> ['C:\\hostedtoolcache\\windows\\Python\\3.12.3\\x64\\python.exe']> is not a valid python or it is missing setuptools`

It also have been reported at the [gvsbuild](https://github.com/wingtk/gvsbuild/issues/1217), a package to help building GTK on Windows.

I managed to resolve it locally but when trying to verify the fix on Github Actions, the build fails.

DjLegolas:
Fixes #265.

Proposed changes:

 - remove the `python3` search from `src/meson.build`.
 - add additional build with `Gobject` integration.


ebassi:
Would it be possible to split it into two separate commits:

1. the change for the Python installation
2. the GitHub workflow update

and rebase it on the main development branch? I've changed the minimum required version of Meson to 1.0.

Also, I'm not sure why the build hasn't started any check; I don't have a Windows machine, so I rely on the CI pipeline to check if it's actually working.

ebassi:
From the build log: `FileNotFoundError: [WinError 2] The system cannot find the file specified`

DjLegolas:
Well, still need to understand why it's failing (it fails to find something...)
In addition, I'll enable the `introspection` tests. But currently it fails on my machine so still investigating it.

PR #267, opened 2024-05-15
Fix GRAPHENE_CHECK_VERSION

elmarco:
The test is backward. (unfortunately, I spread that mistake in virgl too!)


ebassi:
Indeed, nice catch.

PR #268, opened 2024-07-25
meson: add pthreads option

ffontaine:
Allow the user to disable pthreads

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -132,7 +132,6 @@ endif
 
 # Required dependencies
 mathlib = cc.find_library('m', required: false)
-threadlib = dependency('threads')
 
 # Headers
 test_headers = [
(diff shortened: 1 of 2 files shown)


ebassi:
Why? What kind of platform is this, that does not have pthread?

Care to elaborate what is the goal of adding a build time option that I am going to support?

ffontaine:
[uclibc-ng](https://uclibc-ng.org) can be configured without threads support. This configuration can be used by users on some embedded systems (microcontrollers). If this PR is not acceptable, I'll add a dependency on threads for graphene on [buildroot](https://git.buildroot.net/buildroot/tree/package/graphene/Config.in) side.

ebassi:
I very much doubt anybody will ever use graphene on microcontrollers, considering the nature of the project; unless there are microcontrollers with vector instructions somewhere…

> If this PR is not acceptable, I'll add a dependency on threads for graphene on [buildroot](https://git.buildroot.net/buildroot/tree/package/graphene/Config.in) side.

I think it's more appropriate to add a dependency. I definitely do not test Graphene without a platform threading implementation, so I can't ensure it'll ever work.

Thanks for your contribution, in any case.

PR #276, opened 2025-01-20
Replace obsolete isinff(...)

fxing-qnx:
Fixes # N/A

Proposed changes: Replace `isinff(...)` with a combination of `isinf(...)` and `signbit(...)`

Benchmark results: No changes in performance.

Test suite changes: N/A

Almost the same to https://github.com/ebassi/graphene/pull/188, but more aggressive. `isinff(...)` is completely remove from the source tree.
Reason: Mentioned in PR188, `isinff(...)` is both obsolete and non-standard. To maximized portability and suggested by GNU, we should have it replaced.


ebassi:
You included a release commit; please, rebase instead of merging from the main development branch.

Additionally, the reason why #188 was not merged is that it breaks the tests of one of the SIMD implementations.

fxing-qnx:
> You included a release commit; please, rebase instead of merging from the main development branch.
> 
> Additionally, the reason why #188 was not merged is that it breaks the tests of one of the SIMD implementations.

Oh I am so sorry, I will then fix the problem soon

PR #277, opened 2025-01-21
Replace obsolete isinff(...)

fxing-qnx:
Fixes # N/A

Proposed changes: Replace isinff(...) with a combination of isinf(...) and signbit(...)

Benchmark results: No changes in performance.

Test suite changes: N/A

Almost the same to https://github.com/ebassi/graphene/pull/188, but more aggressive. isinff(...) is completely remove from the source tree.
Reason: Mentioned in PR188, isinff(...) is both obsolete and non-standard. To maximized portability and suggested by GNU, we should have it replaced.

PS, Sorry due to our process we need to always fork under tags even if a tag only contains version changes. I have a rebased branch ready for merge now.


ebassi:
> `4/23 graphene / box              ERROR           0.02s   exit status 1`

> `▶  4/23 - box(empty).center() to be in (0, 0, 0) FAIL`

fxing-qnx:
> > `4/23 graphene / box              ERROR           0.02s   exit status 1`
> 
> > `▶  4/23 - box(empty).center() to be in (0, 0, 0) FAIL`

Yeah I saw that, very weird.........

fxing-qnx:
@ebassi `-ffast-math` is set by meson. According to https://gcc.gnu.org/wiki/FloatingPointMath, this flag will also set `-ffnite-math-only` which will violate IEEE's floating-point specification and can potentially cause standard functions like `isinf(...)` to fail. 

The problem on the surface is `isinf(...)` will only work for `double` type instead of `float`. For some reason `isinff(...)` will return the right answer even with those flags.

ebassi:
The MSYS2 build failure in CI is caused by a package rename, and has been fixed in master.

fxing-qnx:
It should now be fine........

fxing-qnx:
@ebassi Are the changes ok now?

PR #283, opened 2025-06-27
Avoid implicit conversion to double

hpjansson:


include/graphene-simd4x4f.h
```diff
@@ -656,7 +656,7 @@ graphene_simd4x4f_init_look_at (graphene_simd4x4f_t *m,
       graphene_simd4f_t tweak_z;
 
       /* up and z_axis are parallel */
-      if (fabs (graphene_simd4f_get_z (up) - 1.0) < FLT_EPSILON)
+      if (fabsf (graphene_simd4f_get_z (up) - 1.f) < FLT_EPSILON)
```
swilmet, line 659: Looks all good and it's still readable. (I'm not a maintainer, but I'm doing a review to help getting this merged).

Everything is float after the change:

- `graphene_simd4f_get_z()` returns a float.
- `1.f` is float.
- `fabsf()` takes a float argument and returns a float.
- `FLT_EPSILON` is afloat.

Tested the patch, it fixes all the warnings that I had (using clang).

I think it would be nice to add an automated/regression test (just for the public `#include <graphene.h>`, not for all the `*.c` files).

swilmet (approved):

